PR #4, opened 2016-01-05
Design overview

JosephSalisbury:


JosephSalisbury:
See https://github.com/giantswarm/architecture/pull/27 for prior discussion

JosephSalisbury:
yo @zyndiecate does this look good to you buddyyyyyy

design/overview.md
```diff
@@ -0,0 +1,145 @@
+# Design Overview
+
+Details the overall design of Formica - a management layer over fleet
+
+- _Formica_ is a genus of ants, see https://en.wikipedia.org/wiki/Formica
+- See overview.png for a more graphical view of this material
```
xh3b4sd, line 6: I like this information.

design/overview.md
```diff
@@ -0,0 +1,145 @@
...
+
+## General Design Overview
+- Service running inside cluster named ```formicad```
+- Command line tool to interact with ```formicad``` named ```formicactl```
+
+### ```formicad```
+- Handles generating unit files from unit file definitions
+- Handles submitting and updating unit files
+- Supports submitting unit file definitions
+- Supports querying against the state of the unit file definition being processed
```
xh3b4sd, line 16: What kind of state information do we want to provide?

design/overview.md
```diff
@@ -0,0 +1,145 @@
...
+- Supports users viewing the generated unit files
+- Fleet is source of truth - no state is held in ```formicad``` that can not be regenerated from fleet
+- Uses a task system with swappable backends
+- No user management, authentication or authorization
```
xh3b4sd, line 20: A future idea might be "registering" users where certificates are generated that then can be used by clients for authentication purposes. So the daemon can be run by only accepting requests with valid authentication.

design/overview.md
```diff
@@ -0,0 +1,145 @@
...
+
+### ```formicactl```
+- Supports sending unit file definitions to ```formicad```
+- Supports querying state of unit file definition being processed
```
xh3b4sd, line 24: And all the common commands
- `up`
- `create`
- `start`
- `stop`
- `delete`
- `update`

design/overview.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Example Flow
+
+### Using command line tool
+- User submits unit file definition to service using command line tool
+- Service validates unit file definition, returns 200
```
xh3b4sd, line 30: This describes the asynchronous way right?

design/overview.md
```diff
@@ -0,0 +1,145 @@
...
+- Service generates unit files from unit file definition
+- Service submits all unit files to fleet cluster
+- Meanwhile, command line tool periodically queries service for unit file definition status, displays status to user
+
+## Shipping
+- Docker image as main deployment artifact
+- Continuous integration with automated unit tests
+- Continuous deployment to build Docker image and push image to Docker Hub
```
xh3b4sd, line 38: Lets see what travic.ci can do for us here.

design/overview.md
```diff
@@ -0,0 +1,145 @@
...
+        },
+        "Service": {
+          "ExecStart": "docker run..."
+        },
+        "X-Fleet": {
+          "Global": true
+        }
+      }
+    }
+  }
```
xh3b4sd: Just to have it done, we should compare JSON with YAML :D

design/overview.md
```diff
@@ -0,0 +1,145 @@
...
+$ cat restart_policy.json
+{
+  "Restart": "on-failure",
+  "RestartSec": "1s",
+  "StartLimitInterval": "300s",
+  "StartLimitBurst": "3",
+  "TimeoutStartSec": "0"
+}
+```
+Units are described by the "systemd" key, whose keys map directly to systemd unit values, allowing new systemd unit values without re-releasing
```
xh3b4sd, line 43: What does re-releasing mean here?

design/overview.md
```diff
@@ -0,0 +1,145 @@
...
+        ...
+    "ambassador.json": {
+        "systemd": {
+        ...
+    },
+    "restart_policy.json": {
+        "Restart": "on-failure",
+        ...
+    }
+}
```
xh3b4sd, line 65: API wise I would like to wrap this into its own block. So we are free to extend the payload with whatever we want without interfering with one specific aspect of the payload.

design/overview.md
JosephSalisbury, line 16: My ideas were largely:
- Has this unit file definition been parsed into unit files?
- How many of the unit files have been submitted?
  As we mentioned in our chat, it's essentially querying the task system for the state of processing

The aim of this querying is to allow clients to see the state of processing, and these are the two main responsibilities of `formicad`

design/overview.md
JosephSalisbury, line 24: We probably also need to be able to specify which `formicad` to talk to (e.g: an `endpoint` command), and `ping` would be useful too

design/overview.md
JosephSalisbury, line 30: Yes, I was under the impression we were going with the asynchronous design

design/overview.md
JosephSalisbury, line 38: Travis.ci has pretty good support for Docker (see: https://docs.travis-ci.com/user/docker/), so CI/CD and building docker images should all be fairly straightforward.

design/overview.md
JosephSalisbury: I'll have a bash at transcribing this into YAML and HCL, so we can have a little look at them all.

design/overview.md
JosephSalisbury, line 43: As in, we don't have a mapping between keys in the unit file definition and systemd unit values in `formicad`.

```
"X-Fleet": {
   "MachineMetadata="role-core=true"
}
```

is translated directly to

```
[X-Fleet]
MachineMetadata="role-core=true"
```

This means we don't have to add support for "X-Whatever" or "SomeOtherValue" with a new release, they're copied directly into the unit file.

design/overview.md
JosephSalisbury, line 65: Yes, I completely agree with this, I'll update

design/overview.md
JosephSalisbury: Correction: There's no automatic JSON -> HCL conversion and I'm lazy, so just YAML incoming.

design/overview.md
xh3b4sd, line 24: +1

design/overview.md
xh3b4sd, line 30: How to identify then that something is done? I think we will need both.

design/overview.md
xh3b4sd, line 38: :+1:

design/overview.md
xh3b4sd, line 43: Got you.

design/overview.md
JosephSalisbury, line 30: Works for me

design/overview.md
puja108, line 20: yeah, this would be in line with a potential "trusted computing" story where we would deliver certificates already in the bootstrapping phase.

design/unit_file_definition/json_example.md
```diff
@@ -0,0 +1,85 @@
+```
+$ cat example.json
+{
+	"units": {
```
zeisss, line 4: a hash here does not provide ordering

hectorj2f:
@JosephSalisbury @zeisss @xh3b4sd Do we still want to move far from Docker compose extensions to allow the definition of apps in a easy way  and out of the box ? I don't know if we want to discuss it here, ...

We know there is quite a lot of users of fleet, I believe having something like Helm for fleet will be very useful. 

Regarding the examples, I understand their shape, but I'd like to imagine the `ambassador`, `lb-registers` being deducted from a docker compose definition :D :D.  This would open the door to a potential catalog of apps to easily deploy in fleemmer using our models.

Any Ideas, thoughts, complaints ?

PR #13, opened 2016-02-11
providing basic fleet client

xh3b4sd:
This PR is supposed to provide a basic fleet client implementation. Higher level scheduling and management needs to be built on top of that later.


fleet/fleet.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// Delete delets a unit on the configured fleet cluster. In case the given unit
+	// is not stopped, UnitRunningError is returned. If there cannot be any unit
+	// found, UnitNotFoundError is returned.
+	Delete(name string) error
+
+	// GetStatus fetches the current status of a unit. If there cannot be any unit
+	// found, UnitNotFoundError is returned.
+	GetStatus(name string) (Status, error)
+}
```
xh3b4sd, line 86: @zeisss @JosephSalisbury I want to have early feedback. What do you think about that?

fleet/fleet.go
```diff
@@ -0,0 +1,87 @@
...
+	FleetStatus containerstates.State
+
+	// HostIP represents the hosts IP where the related unit is running on. If
+	// the unit is not scheduled, HostIP is empty.
+	HostIP net.IP
+}
+
+// Fleet defines the interface a fleet client needs to implement to provide
+// basic operations against a fleet endpoint.
+type Fleet interface {
```
JosephSalisbury, line 66: How about "FleetService"? :D

fleet/fleet.go
```diff
@@ -0,0 +1,87 @@
...
+	// Create schedules a unit on the configured fleet cluster. In case the given
+	// unit already exists, UnitAlreadyExistsError is returned.
+	Create(name, content string) error
+
+	// Start starts a unit on the configured fleet cluster. In case the given
+	// unit is already started, UnitRunningError is returned. If there cannot be
+	// any unit found, UnitNotFoundError is returned.
```
JosephSalisbury: "If the unit cannot be found, UnitNotFoundError is returned."

fleet/fleet.go
xh3b4sd, line 66: The package is called fleet. So it is about having an instance of the implementation the package provides. That is why I named it simply fleet. I prefer the short name `Fleet` though.

fleet/fleet.go
```diff
@@ -0,0 +1,87 @@
...
+
+	return newConfig
+}
+
+type Status struct {
+	// SystemdStatus represents the status within systemd.
+	SystemdStatus containerstates.State
+
+	// FleetStatus represents the status within fleet.
+	FleetStatus containerstates.State
```
zeisss: Desired or Current State?

fleet/fleet.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// HostIP represents the hosts IP where the related unit is running on. If
+	// the unit is not scheduled, HostIP is empty.
+	HostIP net.IP
```
zeisss: > [..], HostIP is empty

Do we want to make this a pointer then? So it can become `nil`?

fleet/fleet.go
zeisss, line 86: I am not sure what this service represents - fleet itself or our wrapper that provides group logic?

fleet/fleet.go
xh3b4sd: I don't know which one is which one right now. This differentiation was ever confusing to me. I oriented by the code comments of other implementations so I named it fleet and systemd status. What do you prefer?

fleet/fleet.go
xh3b4sd: We can do that.

fleet/fleet.go
zeisss: we actually ahve all 3: systemd, fleet current, fleet desired.

fleet/fleet.go
xh3b4sd, line 86: IMO I wrote proper comments even on the package level. Why is that still not clear? This is about fleet. A higher scheduler can use this then and implement grouping.

zeisss, line 86: ack, already forgot that when I came down here. I guess so far this makes sense. We will see when we actually use it.

fleet/fleet.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// Fleet defines the interface a fleet client needs to implement to provide
+// basic operations against a fleet endpoint.
+type Fleet interface {
+	// Create schedules a unit on the configured fleet cluster. In case the given
+	// unit already exists, UnitAlreadyExistsError is returned.
+	Create(name, content string) error
```
zeisss: Do we want to orientate ourself on the fleetctl names? So submit here?

fleet/fleet.go
xh3b4sd: Based on https://www.freedesktop.org/wiki/Software/systemd/dbus/ we have this. 

> LoadState contains a state value that reflects whether the configuration file of this unit has been loaded. The following states are currently defined: loaded, error, masked. loaded indicates that the configuration was successfully loaded. error indicates that the configuration failed to load, the LoadError field (see below) contains information about the cause of this failure. masked indicates that the unit is currently masked out (i.e. symlinked to /dev/null or suchlike). Note that the LoadState is fully orthogonal to the ActiveState (see below) as units without valid loaded configuration might be active (because configuration might have been reloaded at a time where a unit was already active).
> 
> ActiveState contains a state value that reflects whether the unit is currently active or not. The following states are currently defined: active, reloading, inactive, failed, activating, deactivating. active indicates that unit is active (obviously...). reloading indicates that the unit is active and currently reloading its configuration. inactive indicates that it is inactive and the previous run was successful or no previous run has taken place yet. failed indicates that it is inactive and the previous run was not successful (more information about the reason for this is available on the unit type specific interfaces, for example for services in the Result property, see below). activating indicates that the unit has previously been inactive but is currently in the process of entering an active state. Conversely deactivating indicates that the unit is currently in the process of deactivation.
> 
> SubState encodes states of the same state machine that ActiveState covers, but knows more fine-grained states that are unit-type-specific. Where ActiveState only covers six high-level states, SubState covers possibly many more low-level unit-type-specific states that are mapped to the six high-level states. Note that multiple low-level states might map to the same high-level state, but not vice versa. Not all high-level states have low-level counterparts on all unit types. At this point the low-level states are not documented here, and are more likely to be extended later on than the common high-level states explained above.

What about simply providing LoadState, ActiveState and SubState here?

fleet/fleet.go
xh3b4sd: No strong feelings. Feel free to decide.

fleet/fleet.go
zeisss: thats the 3 systemd states that fleet more or less aggregates into one. You can still see them with fleetctl though. See also https://github.com/coreos/fleet/blob/master/Documentation/states.md

If you want to be 100% correct, you have to return all 5 fields, although I would assume we just need ActiveState, FleetDesired, FleetCurrent

xh3b4sd: Based on this documentation I neither see that there are the three specific ones you are talking about nor there are 5 in total. I am still confused. Anyway, `FleetDesired` and `FleetCurrent` makes sense. What is ActiveState? Is it SystemdCurrent?

zeisss: You quoted the documentation of systemd, which gives us 3 states. Plus 2 (desired, current) from fleet, makes 5.
I am confused by your confusion! :dancer: 

> What is ActiveState? 

The state we are most likely interested in.

xh3b4sd: Where does ActiveState come from? Why is it the most interesting one?

xh3b4sd: The UnitState looks like that. 

```
type UnitState struct {
    Hash string `json:"hash,omitempty"`

    MachineID string `json:"machineID,omitempty"`

    Name string `json:"name,omitempty"`

    SystemdActiveState string `json:"systemdActiveState,omitempty"`

    SystemdLoadState string `json:"systemdLoadState,omitempty"`

    SystemdSubState string `json:"systemdSubState,omitempty"`
}
```

So by name it is al systemd related. Where does your `FleetCurrent` and `FleetDesired` come from?

fleet/fleet.go
JosephSalisbury: I'd prefer fleetctl names.

fleet/fleet.go
zeisss: > Why is it the most interesting one?

LoadState just represents whether the unit is loaded or not. Fleetctl itself showed it in the past, but dropped it because its mostly useless. We are interested in whether things are running, which are represented by ActiveState. SubState is just a more detailed version of ActiveState.

> Where does your FleetCurrent and FleetDesired come from?

from the `Unit` itself, you have the UnitState here.

xh3b4sd: Discussed. Will put results in the code.

fleet/fleet.go
xh3b4sd: Done

xh3b4sd:
RFR @JosephSalisbury @zeisss 

CI is failing. Why is that? https://travis-ci.org/giantswarm/formica/builds/108789287

And why do I get emails because of that?

xh3b4sd:
Ok it fails because I need to fetch the fleet dependencies with builder.

zeisss:
that and also: we need to pull dependencies before running the test itself.

xh3b4sd:
RFR again. What do you think?

fleet/fleet.go
```diff
@@ -0,0 +1,220 @@
...
+	// Stop stops a unit on the configured fleet cluster. This is done by
+	// setting the unit's target state to loaded.
+	Stop(name string) error
+
+	// Destroy delets a unit on the configured fleet cluster. This is done by
+	// setting the unit's target state to inactive.
+	Destroy(name string) error
+
+	// GetStatus fetches the current status of a unit. If there cannot be any
+	// unit found, an error that you can identify using IsUnitNotFound is
```
JosephSalisbury: This should read more like "if the unit cannot be found"

xh3b4sd: Done

On Fri, Feb 12, 2016 at 4:26 PM, Joseph Salisbury notifications@github.com
wrote:

> In fleet/fleet.go
> https://github.com/giantswarm/formica/pull/13#discussion_r52754379:
> 
> > -   Submit(name, content string) error
> >   +
> > -   // Start starts a unit on the configured fleet cluster. This is done by
> > -   // setting the unit's target state to launched.
> > -   Start(name string) error
> >   +
> > -   // Stop stops a unit on the configured fleet cluster. This is done by
> > -   // setting the unit's target state to loaded.
> > -   Stop(name string) error
> >   +
> > -   // Destroy delets a unit on the configured fleet cluster. This is done by
> > -   // setting the unit's target state to inactive.
> > -   Destroy(name string) error
> >   +
> > -   // GetStatus fetches the current status of a unit. If there cannot be any
> > -   // unit found, an error that you can identify using IsUnitNotFound is
> 
> This should read more like "if the unit cannot be found"
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/giantswarm/formica/pull/13/files#r52754379.

## 

## 

Tim Schindler
Developer @ Giant Swarm GmbH
tim@giantswarm.io
Web: http://giantswarm.io
Twitter: http://twitter.com/giantswarm
Handelsregister: Amtsgericht Köln, HRB 81484 · Sitz der Gesellschaft: Köln
· Geschäftsführer: Timo Derstappen, Henning Lange, Oliver Thylmann
The latest from Giant Swarm (@giantswarm). Simple Service Orchestration,
built for developers. Cologne, Germany

fleet/error.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"github.com/juju/errgo"
+)
+
+var (
+	maskAny = errgo.MaskFunc(errgo.Any)
+)
+
+var ipNotFoundError = errgo.New("ip not found")
```
zeisss, line 11: so far we have made errors public - any special reason for making this now private?

zeisss:
I like - but we had some learnings in appd that we want retries around fleet operations. do we want to add them later? We also could just copy over the fleet scheduler package from appd as a basic and iterate on that.

fleet/error.go
xh3b4sd, line 11: AFAIK we agreed on no longer making them public. That is why.

xh3b4sd:
My idea was to have the client as minimal as possible. All other logic like grouping of unit files, retries and higher level scheduling actions will be build on top of that. So the client is as clean as possible.

fleet/error.go
zeisss, line 11: Thats new to me, but I don't need to know everything :D

zeisss:
I am not sure thats a wise abstraction, as you may want to retry only certain network operations, not the parsing and other stuff. But we can iterate on that when we need to it. 

LGTM

JosephSalisbury:
I like the idea of having a minimal client, at least to begin with, and iterating if we need to.

xh3b4sd:
Ok, going with this now.

PR #17, opened 2016-02-15
basic controller commands

xh3b4sd:
First basic controller commands. Note this is WIP. 

RFC @zeisss @JosephSalisbury

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,9 @@ test: $(SOURCE) VERSION .gobuild
 	    $(TEST_COMMAND)
 
 lint:
-	go vet -x
+	for source_file in $(SOURCE); do \
+		go vet -x $$source_file ; \
+	done
 
 ci-build: $(SOURCE) VERSION .gobuild
 	echo Building for $(GOOS)/$(GOARCH)
--- a/Makefile
+++ b/Makefile
@@ -4,38 +4,37 @@ BUILD_PATH := $(shell pwd)/.gobuild
 GS_PATH := $(BUILD_PATH)/src/github.com/giantswarm
 GOPATH := $(BUILD_PATH)
 
-BIN := $(PROJECT)
+BIN := $(PROJECT)ctl
 
 VERSION := $(shell cat VERSION)
 COMMIT := $(shell git rev-parse --short HEAD)
 
-.PHONY: all clean test ci-test
+.PHONY: all clean test ci-test deps
 
 SOURCE=$(shell find . -name '*.go')
 
-BUILD_COMMAND=go build -a -o $(BIN)
+BUILD_COMMAND=go build -o formicactl/$(BIN) github.com/giantswarm/formica/$(BIN)
 TEST_COMMAND=go test ./... -cover
 
-all: $(BIN)
+all: formica/$(BIN)
 
 clean:
-	rm -rf $(BUILD_PATH) $(BIN)
+	rm -rf $(BUILD_PATH) formicactl/$(BIN)
 
 .gobuild:
 	@mkdir -p $(GS_PATH)
 	@rm -f $(GS_PATH)/$(PROJECT) && cd "$(GS_PATH)" && ln -s ../../../.. $(PROJECT)
 	#
-	# Fetch private packages first (so `go get` skips them later)
-	# @GOPATH=$(GOPATH) builder go get github.com/spf13/cobra
-	# Pin versions of certain libs
-	@builder get dep -b v0.10.2 https://github.com/coreos/fleet.git $(GOPATH)/src/github.com/coreos/fleet
-	#
-	@GOPATH=$(GOPATH) builder go get github.com/juju/errgo
-	#
-	# Fetch public dependencies via `go get`
-	GOPATH=$(GOPATH) go get -d -v github.com/giantswarm/$(PROJECT)
+	# Fetch and pin packages
+	@builder get dep -b 76516ab4ae194e37aaae9c1f2fa5090553e541f3 https://github.com/coreos/fleet.git $(GOPATH)/src/github.com/coreos/fleet
+	@builder get dep -b 08cceb5d0b5331634b9826762a8fd53b29b86ad8 https://github.com/juju/errgo.git $(GOPATH)/src/github.com/juju/errgo
+	@builder get dep -b 65a708cee0a4424f4e353d031ce440643e312f92 https://github.com/spf13/cobra.git $(GOPATH)/src/github.com/spf13/cobra
+	@builder get dep -b 7f60f83a2c81bc3c3c0d5297f61ddfa68da9d3b7 https://github.com/spf13/pflag.git $(GOPATH)/src/github.com/spf13/pflag
 
-$(BIN): $(SOURCE) VERSION .gobuild
+deps:
+	@${MAKE} -B -s .gobuild
+
+formica/$(BIN): $(SOURCE) VERSION .gobuild
 	@echo Building inside Docker container for $(GOOS)/$(GOARCH)
 	docker run \
 	    --rm \
--- a/Makefile
+++ b/Makefile
@@ -13,13 +13,13 @@ COMMIT := $(shell git rev-parse --short HEAD)
 
 SOURCE=$(shell find . -name '*.go')
 
-BUILD_COMMAND=go build -o formicactl/$(BIN) github.com/giantswarm/formica/$(BIN)
+BUILD_COMMAND=go build -o $(BIN)
 TEST_COMMAND=go test ./... -cover
 
-all: formica/$(BIN)
+all: $(BIN)
 
 clean:
-	rm -rf $(BUILD_PATH) formicactl/$(BIN)
+	rm -rf $(BUILD_PATH) $(BIN)
 
 .gobuild:
 	@mkdir -p $(GS_PATH)
@@ -34,7 +34,7 @@ clean:
 deps:
 	@${MAKE} -B -s .gobuild
 
-formica/$(BIN): $(SOURCE) VERSION .gobuild
+$(BIN): $(SOURCE) VERSION .gobuild
 	@echo Building inside Docker container for $(GOOS)/$(GOARCH)
 	docker run \
 	    --rm \
@@ -46,8 +46,8 @@ formica/$(BIN): $(SOURCE) VERSION .gobuild
 	    golang:1.5 \
 	    $(BUILD_COMMAND)
 
-test:
-	echo Testing inside Docker container for $(GOOS)/$(GOARCH)
+test: $(SOURCE) VERSION .gobuild
+	@echo Testing inside Docker container for $(GOOS)/$(GOARCH)
 	docker run \
 	    --rm \
 	    -v $(shell pwd):/usr/code \
@@ -65,6 +65,6 @@ ci-build: $(SOURCE) VERSION .gobuild
 	echo Building for $(GOOS)/$(GOARCH)
 	$(BUILD_COMMAND)
 
-ci-test:
+ci-test: $(SOURCE) VERSION .gobuild
 	echo Testing for $(GOOS)/$(GOARCH)
 	$(TEST_COMMAND)
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -29,19 +29,19 @@ func DefaultConfig() Config {
 // operations for groups of unit files against a fleet cluster.
 type Controller interface {
 	// Submit schedules a group on the configured fleet cluster. This is done by
-	// setting the units target state to loaded.
+	// setting the state of the units in the group to loaded.
 	Submit(group string) error
 
 	// Start starts a group on the configured fleet cluster. This is done by
-	// setting the units target state to launched.
+	// setting the state of the units in the group to launched.
 	Start(group string) error
 
 	// Stop stops a group on the configured fleet cluster. This is done by
-	// setting the unit's target state to loaded.
+	// setting the state of the units in the group to loaded.
 	Stop(group string) error
 
 	// Destroy delets a group on the configured fleet cluster. This is done by
-	// setting the unit's target state to inactive.
+	// setting the state of the units in the group to inactive.
 	Destroy(group string) error
 
 	// GetStatus fetches the current status of a group. If the unit cannot be
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -1,0 +1,98 @@
+// Package controller implements a controller client providing basic operations against a
+// controller endpoint through controller's HTTP API. Higher level scheduling and
+// management should be built on top of that.
+package controller
+
+import (
+	"github.com/giantswarm/formica/fleet"
+)
+
+type Config struct {
+	Fleet fleet.Fleet
+}
+
+func DefaultConfig() Config {
+	newFleetConfig := fleet.DefaultConfig()
+	newFleet, err := fleet.NewFleet(newFleetConfig)
+	if err != nil {
+		panic(err)
+	}
+
+	newConfig := Config{
+		Fleet: newFleet,
+	}
+
+	return newConfig
+}
+
+// Controller defines the interface a controller needs to implement to provide
+// operations for groups of unit files against a fleet cluster.
+type Controller interface {
+	// Submit schedules a group on the configured fleet cluster. This is done by
+	// setting the units target state to loaded.
+	Submit(group string) error
+
+	// Start starts a group on the configured fleet cluster. This is done by
+	// setting the units target state to launched.
+	Start(group string) error
+
+	// Stop stops a group on the configured fleet cluster. This is done by
+	// setting the unit's target state to loaded.
+	Stop(group string) error
+
+	// Destroy delets a group on the configured fleet cluster. This is done by
+	// setting the unit's target state to inactive.
+	Destroy(group string) error
+
+	// GetStatus fetches the current status of a group. If the unit cannot be
+	// found, an error that you can identify using IsUnitNotFound is returned.
+	GetStatus(group string) (fleet.UnitStatus, error)
+}
+
+func NewController(config Config) Controller {
+	newController := controller{
+		Config: config,
+	}
+
+	return newController
+}
+
+type controller struct {
+	Config
+}
+
+func (c controller) Submit(group string) error {
+	// TODO turn group into unit files
+
+	// TODO submit each unit file
+	unit := ""
+	content := ""
+	err := c.Fleet.Submit(unit, content)
+	if err != nil {
+		return maskAny(err)
+	}
+
+	// TODO retry operations
+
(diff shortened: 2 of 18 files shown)


controller/controller.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+type controller struct {
+	Config
+}
+
+func (c controller) Submit(group string) error {
+	// TODO turn group into unit files
+
+	// TODO submit each unit file
```
xh3b4sd: It is not clear for me right now how this should work or look like. How is the workflow for the enduser supposed to be?

fleet/fleet.go
```diff
@@ -217,3 +256,6 @@ func (f fleet) ipFromUnitState(unitState *schema.UnitState) (net.IP, error) {
 
 	return nil, maskAny(ipNotFoundError)
 }
+
+// TODO
```
zeisss: I wouldn't put this in here

xh3b4sd: Ok I presume you mean the PrintStatus method. Do you have an alternative suggestion or at least reasons why not to put it here? Your blank statement makes it hard to act on it.

controller/controller.go
zeisss: didn't our example in the slides cover this? The user specifies the group name and maybe how many slices to start (and their names), e.g. `appd@{5-10}` which should create slice 5,6,7,8,9,10.

fleet/fleet.go
zeisss: its UI logic, so it should be part of the formicactl package IMO.

controller/controller.go
JosephSalisbury: +1

JosephSalisbury: It might be an idea to make the signature here something like `Submit(group Group) error`, and read the group directory earlier, into a `Group` struct?

formicactl/formicactl.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	newController controller.Controller
+	newFleet      fleet.Fleet
+
+	mainCmd = &cobra.Command{
+		Use:   "formicactl",
+		Short: "orchestrate groups of unit files on Fleet clusters",
+		Long:  "orchestrate groups of unit files on Fleet clusters",
+		Run:   mainRun,
```
JosephSalisbury: Do you need to specify a run for the root command? I don't think you do - I think cobra implicitly will call show help usage if no run is specified - you'll be able to remove `mainRun` too.

fleet/fleet.go
xh3b4sd: It operates on the fleet unit status. The package defines the status and knows how to deal with it. This was my thought. Putting it somewhere else we spread details and knowledge about a package's implementation across the whole project. The caller definitely must be the CLI command side. Functionality dealing with the fleet status should be within the fleet package, IMO. 

What do I miss?

formicactl/create.go
```diff
@@ -0,0 +1,24 @@
...
+	createCmd = &cobra.Command{
+		Use:   "create [group]",
+		Short: "create a group",
+		Long:  "create a group",
+		Run:   createRun,
+	}
+)
+
+func createRun(cmd *cobra.Command, args []string) {
+	if len(args) < 1 || len(args) > 1 {
```
JosephSalisbury: Why not just `if len(args) != 1`?

formicactl/status.go
```diff
@@ -0,0 +1,36 @@
...
+	statusCmd = &cobra.Command{
+		Use:   "status [group]",
+		Short: "status of a group",
+		Long:  "status of a group",
+		Run:   statusRun,
+	}
+)
+
+func statusRun(cmd *cobra.Command, args []string) {
+	if len(args) < 1 || len(args) > 1 {
```
JosephSalisbury: Same again here, I don't get why there's an OR here.

Makefile
```diff
@@ -4,38 +4,37 @@ BUILD_PATH := $(shell pwd)/.gobuild
 GS_PATH := $(BUILD_PATH)/src/github.com/giantswarm
 GOPATH := $(BUILD_PATH)
 
-BIN := $(PROJECT)
+BIN := $(PROJECT)ctl
```
JosephSalisbury, line 7: Can you update the .travis.yml too, to execute `./formicactl --help`

controller/controller.go
xh3b4sd: Sure. This implies that some file system folder is there and contains unit files. So I am bound to a "physical" location when executing the tool. I can bear with that. Does this also apply to all other commands? Is this what we want?

Makefile
```diff
@@ -4,38 +4,37 @@ BUILD_PATH := $(shell pwd)/.gobuild
...
+	@builder get dep -b 76516ab4ae194e37aaae9c1f2fa5090553e541f3 https://github.com/coreos/fleet.git $(GOPATH)/src/github.com/coreos/fleet
+	@builder get dep -b 08cceb5d0b5331634b9826762a8fd53b29b86ad8 https://github.com/juju/errgo.git $(GOPATH)/src/github.com/juju/errgo
+	@builder get dep -b 65a708cee0a4424f4e353d031ce440643e312f92 https://github.com/spf13/cobra.git $(GOPATH)/src/github.com/spf13/cobra
+	@builder get dep -b 7f60f83a2c81bc3c3c0d5297f61ddfa68da9d3b7 https://github.com/spf13/pflag.git $(GOPATH)/src/github.com/spf13/pflag
+
+deps:
+	@${MAKE} -B -s .gobuild
 
-$(BIN): $(SOURCE) VERSION .gobuild
+formica/$(BIN): $(SOURCE) VERSION .gobuild
```
JosephSalisbury: Why are we building inside a directory, if we only have one executable?

controller/controller.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	return newConfig
+}
+
+// Controller defines the interface a controller needs to implement to provide
+// operations for groups of unit files against a fleet cluster.
+type Controller interface {
+	// Submit schedules a group on the configured fleet cluster. This is done by
+	// setting the units target state to loaded.
```
JosephSalisbury: "This is done by setting the target state of the units in the group to `loaded`" :flag-gb:, same below as well

controller/controller.go
zeisss: I would leave reading the file structure to the CLI. I would like to keep this code as portable as possible to be reusable as a library or to be packed into a http service.

zeisss: I guess the signature could be like this:

```
type SubmitRequest struct {
    SliceIDs []string // Contains the IDs to create, leave nil or empty to create a Group Instance
    UnitFiles []struct {
       UnitName string // sthg like "appd@.service", needs to be modified with the sliceID before submitting to fleet
       Content string
   }
}

func Submit(req SubmitRequest) error
```

fleet/fleet.go
JosephSalisbury: Would implementing `String()` on `UnitStatus` work? Then we wouldn't need this function.

Makefile
xh3b4sd: Separation of concerns. Currently we may only have one executable, but I guess we will rather quickly extend the client-only approach. No strong feelings though.

Makefile
xh3b4sd, line 7: I already did.

controller/controller.go
JosephSalisbury: I was thinking something a bit more coarse, like (rough):

```
type Group struct {
   Units []Unit
}

type Unit {
   Name string
   Content string
}
```

fleet/fleet.go
zeisss: > What do I miss?

Separation of concern. This is only used by the cli and has no use elsewhere. Why does the cli even need to know about this fleet package?

> Would implementing String() on UnitStatus work?

Yes, but then you again have cli logic (I want the status to be build like this) in the fleet package.

Makefile
JosephSalisbury: I'd prefer to have it in the top level

controller/controller.go
zeisss: @JosephSalisbury How to express slices etc there?

Makefile
xh3b4sd: Please elaborate how this should look like then. What packages to have? Where to put the binary to?

fleet/fleet.go
xh3b4sd: Even though I think `PrintStatus(UnitStatus) string` belongs to the fleet package I will just move it to the CLI package.

controller/controller.go
xh3b4sd: @zeisss what is the `SliceIDs` list good for?

controller/controller.go
xh3b4sd: Done

Makefile
zeisss: I do a +0.5 for having it in the folder :dancer:

JosephSalisbury: We have `formicactl` as the main build target, which gets built to the top level. I can submit a PR, if you like?

formicactl/create.go
xh3b4sd: Extended legacy. Fixed.

formicactl/status.go
xh3b4sd: Extended legacy. Fixed.

Makefile
xh3b4sd: Please do so.

controller/controller.go
zeisss: to express the slices you want to create? See https://docs.google.com/presentation/d/1N6kOoCJOfF4mHDy7do2cspBIs_5FEWUp2230GFVYaaA/edit#slide=id.g10074f353a_0_25

formicactl/formicactl.go
xh3b4sd: I would like to keep it explicit. This does not hurt. I don't want to hide some magic some library supports nowadays.

controller/controller.go
JosephSalisbury: Good point - in my mental model, groups and group slices are different things, so I'd make them different types. Possibly conversion from Group to GroupSlice types? Uncertain.

formicactl/formicactl.go
JosephSalisbury: Works for me :+1:

controller/controller.go
zeisss: The thing is you give formica the command to create the slice (or group instance), you do not provide it with the object. Thus I would make a separate struct for this here (my `SubmitRequest` struct). 

It may reuse other structs (e.g. the `Unit` you specified) but should be its own type at the root, to be able to easily extend it.

JosephSalisbury: Oh yeah, request struct is good, i'd just like to have types for our other core concepts, as opposed to building it into the request struct (little bit of both approaches)

zeisss: @JosephSalisbury +1

xh3b4sd: Can someone please summarize how the interface now should look like? I cannot follow your discussion.

zeisss:
```
type Unit struct {
      UnitName string // sthg like "appd@.service", needs to be modified with the sliceID before submitting to fleet
       Content string

}
type SubmitRequest struct {
     // Contains the IDs to create, leave nil or empty to create a Group Instance
     // "1", "first", "whatever", "5", etc.
    SliceIDs []string 
    UnitFiles []Unit
}

func Submit(req SubmitRequest) error
```

I think this was the latest version.  How the other functions look like needs to be discussed later I guess.

zeisss: sliceIDs come from the CLI. 

```
$ formicactl submit appd@{1-10}  #1
$ formicactl submit appd@first     #2
$ formicactl submit fleet-unit-replicator #3
```
1. should create 10 slices, "1", "2", to "10"
2. should create 1 slice "first"
3. should create a group instance only

zeisss: as @JosephSalisbury already suggested we may want to have a separate type for the slice-id as well.

xh3b4sd: Done

fleet/fleet.go
xh3b4sd: Done

xh3b4sd:
I implemented your feedback. Please have another look. Need to test and extend tomorrow.

controller/controller.go
```diff
@@ -0,0 +1,191 @@
...
+// Controller defines the interface a controller needs to implement to provide
+// operations for groups of unit files against a fleet cluster.
+type Controller interface {
+	// Submit schedules a group on the configured fleet cluster. This is done by
+	// setting the state of the units in the group to loaded.
+	Submit(req Request) error
+
+	// Start starts a group on the configured fleet cluster. This is done by
+	// setting the state of the units in the group to launched.
+	Start(req Request) error
```
zeisss, line 44: do we really need all unit files for each operation here?

controller/controller.go
```diff
@@ -0,0 +1,191 @@
...
+	// setting the state of the units in the group to loaded.
+	Stop(req Request) error
+
+	// Destroy delets a group on the configured fleet cluster. This is done by
+	// setting the state of the units in the group to inactive.
+	Destroy(req Request) error
+
+	// GetStatus fetches the current status of a group. If the unit cannot be
+	// found, an error that you can identify using IsUnitNotFound is returned.
+	GetStatus(req Request) ([]fleet.UnitStatus, error)
```
zeisss, line 56: I think for `GetStatus()` we only need the group name and the sliceIDs.

controller/controller_test.go
```diff
@@ -0,0 +1,102 @@
+package controller
```
zeisss, line 1: Tests! :cake:

controller/controller_test.go
```diff
@@ -0,0 +1,102 @@
...
+						Content: "some unit content",
+					},
+					{
+						Name:    "unit@foo.service",
+						Content: "some unit content",
+					},
+				},
+			},
+		},
+	}
```
zeisss, line 127: I am missing a test for a group with multiple files.

controller/controller.go
xh3b4sd, line 44: No we don't. I just thought that it is nearly the same interface. I can create a new request type without the content. No strong feelings.

controller/controller.go
xh3b4sd, line 56: The same applies to all other commands but `Submit`. See my comment above.

controller/controller_test.go
xh3b4sd, line 127: Done

xh3b4sd:
@zeisss @JosephSalisbury I would like to merge this. My Plan is to keep the PRs smaller. Next one would be #20 to test the file system specific actions.

controller/controller.go
```diff
@@ -0,0 +1,191 @@
...
+package controller
+
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/giantswarm/formica/fleet"
+)
+
+type Config struct {
```
JosephSalisbury, line 15: This could use godoc

controller/controller.go
```diff
@@ -0,0 +1,191 @@
...
+	Fleet fleet.Fleet
+}
+
+func DefaultConfig() Config {
```
JosephSalisbury, line 21: Same here

controller/controller.go
```diff
@@ -0,0 +1,191 @@
...
+	// Destroy delets a group on the configured fleet cluster. This is done by
+	// setting the state of the units in the group to inactive.
+	Destroy(req Request) error
+
+	// GetStatus fetches the current status of a group. If the unit cannot be
+	// found, an error that you can identify using IsUnitNotFound is returned.
+	GetStatus(req Request) ([]fleet.UnitStatus, error)
+}
+
+func NewController(config Config) Controller {
```
JosephSalisbury, line 66: godoc here too

JosephSalisbury:
Could use godoc strings in a fair few places

JosephSalisbury:
Which go vet should have found, I'll look into it

controller/controller.go
xh3b4sd, line 15: Done

controller/controller.go
xh3b4sd, line 21: Done

controller/controller.go
xh3b4sd, line 66: Done

controller/controller.go
```diff
@@ -0,0 +1,219 @@
...
+			newUnit := unit
+			newUnit.Name = unitExp.ReplaceAllString(newUnit.Name, fmt.Sprintf("@%s.service", sliceID))
+			newRequest.Units = append(newRequest.Units, newUnit)
+		}
+	}
+
+	return newRequest, nil
+}
+
+func (c controller) Submit(req Request) error {
```
JosephSalisbury, line 128: godoc

controller/controller.go
```diff
@@ -0,0 +1,219 @@
...
+			return maskAny(err)
+		}
+	}
+
+	// TODO retry operations
+
+	return nil
+}
+
+func (c controller) Start(req Request) error {
```
JosephSalisbury, line 146: godoc

controller/controller.go
```diff
@@ -0,0 +1,219 @@
...
+			return maskAny(err)
+		}
+	}
+
+	// TODO retry operations
+
+	return nil
+}
+
+func (c controller) Stop(req Request) error {
```
JosephSalisbury, line 164: godoc

controller/controller.go
```diff
@@ -0,0 +1,219 @@
...
+			return maskAny(err)
+		}
+	}
+
+	// TODO retry operations
+
+	return nil
+}
+
+func (c controller) Destroy(req Request) error {
```
JosephSalisbury, line 182: godoc

controller/controller.go
```diff
@@ -0,0 +1,219 @@
...
+			return maskAny(err)
+		}
+	}
+
+	// TODO retry operations
+
+	return nil
+}
+
+func (c controller) GetStatus(req Request) ([]fleet.UnitStatus, error) {
```
JosephSalisbury, line 200: godoc

fleet/fleet.go
```diff
@@ -86,6 +87,44 @@ type Fleet interface {
 }
 
 func NewFleet(config Config) (Fleet, error) {
```
JosephSalisbury, line 99: godoc

xh3b4sd, line 99: Done

xh3b4sd:
As discussed. I will not add comments on the private stuff. It implements an interface which is already documented. There are no special details worth documenting for the specific implementation.

JosephSalisbury:
+1

xh3b4sd:
Fishing for LGTM

JosephSalisbury:
_and you shall have it_ <3

LGTM

PR #18, opened 2016-02-16
Fiddling with basic-cli-commands branch

JosephSalisbury:
I've done a few things here:
- Modified Makefile to build `formicactl` only. We're only building one binary, and I don't like sticking it in a directory. Let's keep things simple.
- Fixed up requirements for some targets - this basically means we don't have to worry about building before testing on travis.
- Moved `formicactl` package to `cli` - largely clerical, I want the binary to be called `formicactl`, and I think `cli` as a package name is somewhat more common.
- `formica.go` to `main.go` - felt weird having it called `formica.go` if we're building `formicactl`, but also felt weird calling it `formicactl.go` if the repository is called `formica`. Very close to golfing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,6 @@ before_install:
 
 script:
 - make lint
-- make ci-build
 - make ci-test
-- ./formicactl/formicactl
+- make ci-build
+- ./formicactl
--- a/formica.go
+++ b/formica.go
@@ -1,3 +1,0 @@
-package main
-
-func main() {}
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
-formica
-formicactl/formicactl
+formicactl
 .gobuild/
--- a/Makefile
+++ b/Makefile
@@ -13,13 +13,13 @@ COMMIT := $(shell git rev-parse --short HEAD)
 
 SOURCE=$(shell find . -name '*.go')
 
-BUILD_COMMAND=go build -o formicactl/$(BIN) github.com/giantswarm/formica/$(BIN)
+BUILD_COMMAND=go build -o $(BIN)
 TEST_COMMAND=go test ./... -cover
 
-all: formica/$(BIN)
+all: $(BIN)
 
 clean:
-	rm -rf $(BUILD_PATH) formicactl/$(BIN)
+	rm -rf $(BUILD_PATH) $(BIN)
 
 .gobuild:
 	@mkdir -p $(GS_PATH)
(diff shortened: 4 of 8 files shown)


.travis.yml
```diff
@@ -15,6 +15,6 @@ before_install:
 
 script:
 - make lint
-- make ci-build
 - make ci-test
-- ./formicactl/formicactl
+- make ci-build
```
JosephSalisbury, line 19: I've put testing first because it feels right in my head, no other reason :dancer:

JosephSalisbury:
ping @xh3b4sd @zeisss have at it

formica.go
```diff
@@ -1,3 +0,0 @@
-package main
```
zeisss, line 1: . @teemow was a loud advocate of having the main function in a file named after the tool, so actually `formicactl.go`. Is reasoning was that other tools like etcd, fleet etc do it like this too. Although nowadays fleet has a `fleetctl` and `fleetd` folder for the two binaries it creates. And in there is a file `<name>.go`.

zeisss:
I prefer `formicactl/` over `cli/`.

formica.go
JosephSalisbury, line 1: No issue with `formicactl.go` over `main.go`, if you lean that way.

JosephSalisbury:
I prefer having the binary in the top-level though, as `formicactl` :(

zeisss:
@JosephSalisbury then put a `formicactl.go` on root and the commands in a `formicactl/` folder?

formica.go
teemow, line 1: lol. i adapted myself the other way around. already started using `main.go` in moa/onsho too :)

so no strong feelings. we should finally decide and put this in the coding guidelines.

https://github.com/giantswarm/onsho/blob/master/main.go

.travis.yml
xh3b4sd, line 19: I needed to change that order recently because the CI failed. When the dependencies are properly set within the Makefile, all good. No strong feelings.

xh3b4sd:
No strong feelings here.

JosephSalisbury:
@zeisss - issue with that, is that we'd be naming the folder `formicactl/`, and the binary would also be `formicactl`, which then has issues

zeisss:
> which then has issues

you mean the build process has issues? Thats why we have `bin/` folders :P but I see your point. Lets go with the current iteration. We can change this later if we actually have multiple binaries. (I would expect us to actually implement just a `formicactl server` command instead, but we will see)

zeisss:
lgtm

PR #19, opened 2016-02-17
added codecov support

xh3b4sd:
This is to support codecov integration.


xh3b4sd:
RFR @JosephSalisbury @zeisss 

Note that the coverage is 0 because there are no tests in this branch. We should write tests though soon.

zeisss:
LGTM

PR #20, opened 2016-02-17
added basic file system implementation

xh3b4sd:
This is to provide file system implementations we can abstract and test against.


xh3b4sd:
RFR @JosephSalisbury @zeisss

file-system/spec/spec.go
```diff
@@ -0,0 +1,19 @@
...
+// implementations.
+package filesystemspec
+
+import (
+	"os"
+)
+
+// FileSystem implements file system operations.
+type FileSystem interface {
+	// ReadDir implements os.ReadDir.
```
zeisss: this godoc is misleading, as the references stdlib things are functions, not interfaces. How about "ReadDir is the equivalent to os.ReadDir"

file-system/fake/file.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	"bytes"
+	"errors"
+	"os"
+	"time"
+)
+
+var (
+	ErrFileOperationNotPermitted = errors.New("fake file does not exist and therefore this operation is not permitted")
```
zeisss: do we want to put these into the `error.go` file? (or even into the spec?)

zeisss: If we would use `errgo.Newf()` here, we get the codeline where the error is defined too

file-system/fake/fake.go
```diff
@@ -0,0 +1,77 @@
...
+
+	return nil, maskAny(pathErr)
+}
+
+func (f *fake) WriteFile(filename string, bytes []byte, perm os.FileMode) error {
+	dir := filepath.Dir(filename)
+	base := filepath.Base(filename)
+
+	if _, ok := f.Storages[dir]; !ok {
+		f.Storages[dir] = map[string][]byte{}
```
zeisss, line 72: so this has the effect that the folder of `WriteFile` is silently created, which I like. I don't think we have the same behavior in `real.WriteFile()` though.

xh3b4sd, line 72: There is a slight difference between the implementations yes. All I want is the abstraction and the tests. Both things are accomplished. I am fine with that.

file-system/spec/spec.go
xh3b4sd: Fixed

file-system/fake/file.go
xh3b4sd: Copy/paste. Removed errors since they are not used.

xh3b4sd:
I would like to merge this. Feedback please.

file-system/fake/fake.go
zeisss, line 72: But having different behaviors in the implementation makes your tests useless if in production we use a different backend that behaves different.

file-system/fake/file.go
```diff
@@ -0,0 +1,104 @@
...
+// Fake represents are readable in memory version of os.File which
+// can be used for testing.
+type file struct {
+	Name    string
+	Mode    os.FileMode
+	ModTime time.Time
+	Buffer  *bytes.Reader
+}
+
+// newFile creates a new file instances based on a file name
```
zeisss: s/instances/instance/

zeisss:
found a typo, still LGTM

file-system/fake/fake.go
xh3b4sd, line 72: Is this really critical to you? What do you suggest to do then?

file-system/fake/file.go
xh3b4sd: Done

file-system/fake/fake.go
zeisss, line 72: Not critical for now, you got a LGTM :cake: 

Still, I would prefer either a) add a comment to the Spec and just state that the behavior is optional or b) add a `os.MkdirAll()` call to the real implementation to align it.

xh3b4sd, line 72: Done

PR #21, opened 2016-02-17
Fix go vet

JosephSalisbury:
This actually runs go vet over all source files.

Note: It won't pick up missing godoc - go lint can do this, but it has way too many false positives to be part of the build process, we're going to have to keep an eye on documentation.


JosephSalisbury:
@xh3b4sd RFR

xh3b4sd:
Besides the ugly Makefile magic, LGTM.

PR #22, opened 2016-02-17
fleet: Added fleet-client mock and simple

zeisss:
This PR tries to improve the test coverage of our `fleet/fleet.go` client by providing it with a mock that records the performed calls.

Future work can include configuring responses and ensuring calls were done in a specific order.


zeisss:
@xh3b4sd Wanna take a look? So far the tests only checks if backend calls were made, I do not return fake results, thus I haven't tested the `Status()` function. I would do that as a next step.

xh3b4sd:
LGTM so far. Go ahead.

PR #23, opened 2016-02-18
handle slice expansion

xh3b4sd:
I checked `fleetctl` and there is no range expansion for the slices. So `fleetctl start unit@{1..3}` is supported via the executing shell. The result is this `{"units/simple@1.service", "units/simple@2.service", "units/simple@3.service"}`. We now need to handle the arguments properly.

RFR @JosephSalisbury @zeisss

### DIFF
--- a/cli/common.go
+++ b/cli/common.go
@@ -65,8 +65,8 @@ func createSliceIDs(slices []string) ([]string, error) {
 		}
 		if len(found) == 0 {
 			// When there is no slice expression "@" given, we are dealing with a
-			// normal dirname, so we take the slice as it is.
-			found = []string{slice}
+			// normal dirname, so there is no slice ID required.
+			continue
 		}
 		sliceIDs = append(sliceIDs, atExp.ReplaceAllString(found[0], ""))
 	}
--- a/cli/common.go
+++ b/cli/common.go
@@ -10,8 +10,9 @@ import (
 
 var groupExp = regexp.MustCompile("@(.*)")
 
-func readUnitFiles(dirName string) (map[string]string, error) {
-	dirName = groupExp.ReplaceAllString(dirName, "")
+func readUnitFiles(slices []string) (map[string]string, error) {
+	slice := slices[0]
+	dirName := groupExp.ReplaceAllString(slice, "")
 
 	fileInfos, err := newFileSystem.ReadDir(dirName)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


zeisss:
> {"units/simple@1.service", "units/simple@2.service", "units/simple@3.service"}

So the shell also expands the subfolders?

xh3b4sd:
The shell expands arguments. Arguments are simple strings. The example output in the description is only a example I played around with.

xh3b4sd:
I added documentation. Please have another look.

docs/slice_expansion.md
```diff
@@ -0,0 +1,36 @@
...
+myapp@1
+myapp@2
+myapp@3
+```
+
+Formica takes these arguments and looks up the directory `myapp`. Then the
+slice IDs `1`, `2` and `3` are used to create the unit file names accordingly.
+So lets imagine there are this two unit files within `myapp`.
+```
+some_unit_name@.service
```
zeisss: I would add the app as a prefix here, although its not 100% important to explain this feature, it might confuse people.

xh3b4sd: Done

cli/common.go
```diff
@@ -62,7 +93,7 @@ func createRequest(dirName string) (controller.Request, error) {
 		req.Units = append(req.Units, controller.Unit{Name: name, Content: content})
 	}
 
-	sliceIDs, err := createSliceIDs(dirName)
+	sliceIDs, err := createSliceIDs(slices)
```
zeisss, line 96: do we maybe want to return the actual groupname from `createSliceIDs` too? Our `controller.Request` is missing that, although I think we should add it there.

cli/common_test.go
```diff
@@ -63,6 +63,29 @@ func Test_Common_createRequest(t *testing.T) {
...
+			Setup: []testFileSystemSetup{
+				{
+					FileName:    "dirname/dirname_unit@.service",
+					FileContent: []byte("some unit content"),
+					FilePerm:    os.FileMode(0644),
+				},
+			},
+			Input: []string{"dirname@1"},
+			Expected: controller.Request{
+				SliceIDs: []string{"1", "foo", "5"},
```
zeisss, line 80: I don't get this test - where is `foo` and `5` coming from?

cli/common_test.go
```diff
@@ -63,6 +63,29 @@ func Test_Common_createRequest(t *testing.T) {
 				},
 			},
 		},
+
+		// This test ensures that loading a single unit from a directory with the
+		// slice expression "@1", "@foo" and "@5" results in the expected
+		// controller request.
+		{
+			Setup: []testFileSystemSetup{
```
zeisss, line 71: FYI: we could return this standard test setup into a function which we just repeat here:

```
{ 
  Setup: GivenSimpleSingleUnitFileSystemSetup(),
  [..]
```

cli/common_test.go
```diff
@@ -30,7 +30,7 @@ func Test_Common_createRequest(t *testing.T) {
 					FilePerm:    os.FileMode(0644),
 				},
 			},
-			Input: "dirname",
+			Input: []string{"dirname"},
```
zeisss, line 33: I think the `Setup` for this test shouldn't have a `@` in the unitfilename.

zeisss:
a few minor things about the tests, but LGTM

cli/common.go
xh3b4sd, line 96: Why should we do that? For what is it needed?

cli/common_test.go
xh3b4sd, line 33: Fixed.

cli/common_test.go
xh3b4sd, line 71: Currently not a big fan of that.

cli/common.go
zeisss, line 96: mh, its probably indeed useless, but to me it feels more correct. Lets assume this is the call:

```
err := controller.Submit(controller.SubmitRequest{
  SliceIDs: []string{"1", "2", "3"},
  Units: []controller.Unit{
     controller.Unit{"appd-main@.service", "..."},
     controller.Unit{"appd-lb-register@.service", "..."},
     controller.Unit{"appd-ambassador@.service", "..."},
  }
})
```

to my eye it just feels weird to not have a `Name` field in there. Also I would probably expect the controller to double check the provided arguments (do the units start with `Name`, if we have slices, do the units contain a `@` etc.), so then its needed. 

But again: Currently its not needed and this is purely my mind wishing this for sugar reasons.

cli/common_test.go
xh3b4sd, line 80: The input should provide that. The fact that the test works anyway is bad. Found the bug. Fixed the test.

cli/common.go
xh3b4sd, line 96: My sugar reasoning was against having this stuff in the cli package. Anyway, lets see where we are going.

PR #26, opened 2016-02-19
Fleet status tests

zeisss:
- Replaces the custom mock with a library which also provides to define number of calls, not calls etc.
- Add test for simple successfull `fleet.GetStatus` call
- Refactors `fleet` to have a `GetStatusWithMatcher` func so we don't need a regexp everywhere

Will have merge conflicts after https://github.com/giantswarm/formica/pull/25

### DIFF
--- a/fleet/fleet.go
+++ b/fleet/fleet.go
@@ -217,7 +217,7 @@ func (f fleet) Destroy(name string) error {
 func (f fleet) GetStatus(name string) (UnitStatus, error) {
 	matcher := func(s string) bool {
 		return name == s
-	}	
+	}
 	unitStatus, err := f.GetStatusWithMatcher(matcher)
 	if err != nil {
 		return UnitStatus{}, maskAny(err)
--- a/fleet/fleet.go
+++ b/fleet/fleet.go
@@ -215,12 +215,10 @@ func (f fleet) Destroy(name string) error {
 }
 
 func (f fleet) GetStatus(name string) (UnitStatus, error) {
-	exp, err := regexp.Compile(fmt.Sprintf("^%s$", name))
-	if err != nil {
-		return UnitStatus{}, maskAny(err)
-	}
-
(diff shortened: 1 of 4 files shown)


zeisss:
ping @giantswarm/cena

zeisss:
I would like to merge #27 beforehand

xh3b4sd:
I have no strong feelings on this.

fleet/fleet.go
```diff
@@ -234,6 +232,13 @@ func (f fleet) GetStatus(name string) (UnitStatus, error) {
 }
 
 func (f fleet) GetStatusWithExpression(exp *regexp.Regexp) ([]UnitStatus, error) {
+	status, err := f.GetStatusWithMatcher(exp.MatchString)
+	return status, maskAny(err)
+}
+
+// GetStatusWithMatcher returns a UnitStatus for each unit where the given matcher
```
JosephSalisbury, line 239: This is slightly misleading, if I understand correctly, as we can return `[]UnitStatus`, not just `UnitStatus`

JosephSalisbury, line 239: Well, not misleading, maybe it's just me - I'd go with `GetStatusWithMatcher returns a list of UnitStatus for each unit where the given matcher returns true`

JosephSalisbury:
Yup, LGTM excluding small documentation comment

fleet/fleet.go
zeisss, line 239: doesn't the "for each unit" imply that it is a list?

JosephSalisbury, line 239: Hence my hesitance :D - I think when I quickly scan it, I see `returns a UnitStatus`, and get slightly confused

zeisss:
Due to rebase and merges this ended up as a86f95686571534f202b4bdef724e86a793876c1 in master.

PR #30, opened 2016-02-22
group status

xh3b4sd:
``` bash
# When all states are equal they are grouped.
core@core-01 ~ $ ./formicactl status units@{1..3}
Group    Units  FDState  FCState  SAState   IP              Machine

units@1  *      loaded   loaded   inactive  192.168.33.101  49ecbf33492c4beaaa5cd9fe72ee881a
units@2  *      loaded   loaded   inactive  192.168.33.101  49ecbf33492c4beaaa5cd9fe72ee881a
units@3  *      loaded   loaded   inactive  192.168.33.101  49ecbf33492c4beaaa5cd9fe72ee881a

# When states differ within a group the state is expanded
core@core-01 ~ $ ./formicactl status units@{1..3}
Group    Units                FDState   FCState   SAState   IP              Machine

units@1  *                    loaded    loaded    inactive  192.168.33.101  49ecbf33492c4beaaa5cd9fe72ee881a
units@2  units_one@2.service  launched  launched  active    192.168.33.101  49ecbf33492c4beaaa5cd9fe72ee881a
units@2  units_two@2.service  loaded    loaded    inactive  192.168.33.101  49ecbf33492c4beaaa5cd9fe72ee881a
units@3  *                    loaded    loaded    inactive  192.168.33.101  49ecbf33492c4beaaa5cd9fe72ee881a
```


fleet/fleet.go
```diff
@@ -74,6 +75,93 @@ type UnitStatus struct {
...
 	// Name represents the unit file name.
 	Name string
+
+	// Slice represents the slice expression. E.g. @1, or @foo, or @5., etc..
+	Slice string
+}
+
+type UnitStatusList []UnitStatus
+
+func (usl UnitStatusList) Group() ([]UnitStatus, error) {
```
zeisss: Can we put this logic into the controller?

xh3b4sd: What are your thoughts behind this?

zeisss: this feels like business logic that belongs to the controller, not the way we talk to fleet. It's part of the higher level stuff that does not belong in here IMHO.

xh3b4sd: Ok, done.

xh3b4sd:
:fishing_pole_and_fish: for LGTM

xh3b4sd:
@zeisss @JosephSalisbury

controller/status.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	return newList, ID, nil
+}
+
+func allStatesEqual(usl []fleet.UnitStatus) bool {
+	for _, us1 := range usl {
+		for _, us2 := range usl {
+			if us1.Current != us2.Current {
+				return false
```
zeisss, line 65: looks like we only have test cases for the `return true`  case at the end. Do we want tests for the other cases in here?

zeisss:
LGTM, although I would prefer a test for the "bad" case of the aggregation logic too.

controller/status.go
xh3b4sd, line 65: Done

Issue #28, opened 2016-02-22
Submit should block until units are submitted


zeisss:
```
core@core-01 ~ $ ./formicactl submit demo@1 demo@2
core@core-01 ~ $ ./formicactl status demo
Group  Units  FDState  FCState  SAState  IP  Machine

core@core-01 ~ $ ./formicactl status demo
Group  Units                FDState  FCState  SAState   IP            Machine

demo   demo-main@1.service  loaded   loaded   inactive  172.17.8.101  4ea5734c9d6347829d9236502b79c6ca
demo   demo-main@2.service  loaded   loaded   inactive  172.17.8.102  52f3c1e0193243d8aedbf54052f83c84
```

afaik we said we want to wait for the status to be reached and have and have `--no-block` option.

zeisss:
we are already blocking now.

PR #32, opened 2016-02-23
formicactl: Add stop and destroy command

zeisss:
This PR adds the stop and destroy command. Its actually a copy and paste of the `formicactl start` command.

I guess #28 becomes more interesting with this, as a destroy on a running group may leave artifacts.

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -312,6 +312,70 @@ func TestController_Start(t *testing.T) {
 	mock.AssertExpectationsForObjects(t, fleetMock.Mock)
 }
 
+
+func TestController_Destroy(t *testing.T) {
+	RegisterTestingT(t)
+
+	// Mocks
+	controller, fleetMock := GivenController()
+	fleetMock.On("GetStatusWithMatcher", mock.AnythingOfType("func(string) bool")).Return(
+		[]fleet.UnitStatus{
+			{
+				Name: "test-main@1.service",
+			},
+			{
+				Name: "test-sidekick@1.service",
+			},
+		},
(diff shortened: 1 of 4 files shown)


zeisss:
ping @giantswarm/cena

controller/controller_test.go
```diff
@@ -312,6 +312,70 @@ func TestController_Start(t *testing.T) {
...
+		},
+		nil,
+	).Once()
+	fleetMock.On("Destroy", "test-main@1.service").Return(nil).Once()
+	fleetMock.On("Destroy", "test-sidekick@1.service").Return(nil).Once()
+
+	// Execute test
+	req := Request{
+		Group:    "test",
+		SliceIDs: []string{"1"},
```
xh3b4sd, line 338: It looks still weird without the unit name.

zeisss, line 338: What unit name?

xh3b4sd, line 338: Discussed. The actual reason why there is no unit name required and why no `Unit` is given, is the fact that `Destroy` and certain calls do not require the unit name at all. The commands lookup the groups and fetch unit names by the given slice ids to operate on what they found.

zeisss, line 338: We also talked about that we maybe want a separate Request struct for the Submit case so the `Unit` field is not always there. See https://github.com/giantswarm/giantswarm/issues/527

zeisss:
@giantswarm/cena this PR decreases the coverage because we have no CLI tests. I would not care.

xh3b4sd:
LGTM

PR #34, opened 2016-02-24
added basic task concept

xh3b4sd:
This adds a basic task concept to be able to make synchronous actions asynchronously. Note that this for now only is about the task concept. The following step would be to, e.g. wait for something to be started. 

``` bash
# start something that is not there does nor work
core@core-01 ~ $ ./formicactl start units@{1..3}
Failed to start 3 slices for group 'units': [1 2 3]. (unit not found)

# find something that is not there does not work
core@core-01 ~ $ ./formicactl status units@{1..3}
Failed to find 3 slices for group 'units': [1 2 3].

# submit works then
core@core-01 ~ $ ./formicactl submit units@{1..3}
Succeeded to submit 3 slices for group 'units': [1 2 3].

# status works then
core@core-01 ~ $ ./formicactl status units@{1..3}
Group    Units  FDState  FCState  SAState   IP              Machine

units@1  *      loaded   loaded   inactive  192.168.33.103  027494415d6848078626e57fff3cf801
units@2  *      loaded   loaded   inactive  192.168.33.101  49ecbf33492c4beaaa5cd9fe72ee881a
units@3  *      loaded   loaded   inactive  192.168.33.103  027494415d6848078626e57fff3cf801

# start works then
core@core-01 ~ $ ./formicactl start units@{1..3}
Succeeded to start 3 slices for group 'units': [1 2 3].

# all started
core@core-01 ~ $ ./formicactl status units@{1..3}
Group    Units  FDState   FCState   SAState  IP              Machine

units@1  *      launched  launched  active   192.168.33.103  027494415d6848078626e57fff3cf801
units@2  *      launched  launched  active   192.168.33.101  49ecbf33492c4beaaa5cd9fe72ee881a
units@3  *      launched  launched  active   192.168.33.103  027494415d6848078626e57fff3cf801

# destroy works then
core@core-01 ~ $ ./formicactl destroy units@{1..3}
Succeeded to destroy 3 slices for group 'units': [1 2 3].

# nothing here anymore
core@core-01 ~ $ ./formicactl status units@{1..3}
Failed to find 3 slices for group 'units': [1 2 3].

# nothing here anymore
core@core-01 ~ $ ./formicactl destroy units@{1..3}
Failed to destroy 3 slices for group 'units': [1 2 3]. (unit not found)
```

RFC @zeisss @JosephSalisbury

### DIFF
--- a/cli/formicactl.go
+++ b/cli/formicactl.go
@@ -11,7 +11,6 @@ import (
 	"github.com/giantswarm/formica/file-system/real"
 	"github.com/giantswarm/formica/file-system/spec"
 	"github.com/giantswarm/formica/fleet"
-	"github.com/giantswarm/formica/task"
 )
 
 var (
@@ -24,7 +23,6 @@ var (
 	newController controller.Controller
 	newFileSystem filesystemspec.FileSystem
 	newFleet      fleet.Fleet
-	newTask       task.Task
 
 	MainCmd = &cobra.Command{
 		Use:   "formicactl",
@@ -52,16 +50,13 @@ var (
 			newController = controller.NewController(newControllerConfig)
 
 			newFileSystem = filesystemreal.NewFileSystem()
-
-			newTaskConfig := task.DefaultTaskConfig()
-			newTask = task.NewTask(newTaskConfig)
 		},
 	}
 )
 
 func init() {
 	MainCmd.PersistentFlags().StringVar(&globalFlags.FleetEndpoint, "fleet-endpoint", "unix:///var/run/fleet.sock", "endpoint used to connect to fleet")
-	MainCmd.PersistentFlags().BoolVar(&globalFlags.NoBlock, "no-block", false, "block on syncronous actions or not")
+	MainCmd.PersistentFlags().BoolVar(&globalFlags.NoBlock, "no-block", false, "block on synchronous actions or not")
 	MainCmd.PersistentFlags().BoolVarP(&globalFlags.Verbose, "verbose", "v", false, "verbose output or not")
 
 	MainCmd.AddCommand(submitCmd)
--- a/cli/formicactl.go
+++ b/cli/formicactl.go
@@ -11,17 +11,20 @@ import (
 	"github.com/giantswarm/formica/file-system/real"
 	"github.com/giantswarm/formica/file-system/spec"
 	"github.com/giantswarm/formica/fleet"
+	"github.com/giantswarm/formica/task"
 )
 
 var (
 	globalFlags struct {
 		FleetEndpoint string
+		NoBlock       bool
 		Verbose       bool
 	}
 
 	newController controller.Controller
 	newFileSystem filesystemspec.FileSystem
 	newFleet      fleet.Fleet
+	newTask       task.Task
 
 	MainCmd = &cobra.Command{
 		Use:   "formicactl",
@@ -49,12 +52,16 @@ var (
 			newController = controller.NewController(newControllerConfig)
 
 			newFileSystem = filesystemreal.NewFileSystem()
+
+			newTaskConfig := task.DefaultTaskConfig()
+			newTask = task.NewTask(newTaskConfig)
 		},
 	}
 )
 
 func init() {
 	MainCmd.PersistentFlags().StringVar(&globalFlags.FleetEndpoint, "fleet-endpoint", "unix:///var/run/fleet.sock", "endpoint used to connect to fleet")
+	MainCmd.PersistentFlags().BoolVar(&globalFlags.NoBlock, "no-block", false, "block on syncronous actions or not")
 	MainCmd.PersistentFlags().BoolVarP(&globalFlags.Verbose, "verbose", "v", false, "verbose output or not")
 
 	MainCmd.AddCommand(submitCmd)
--- a/cli/start.go
+++ b/cli/start.go
@@ -5,6 +5,8 @@ import (
 	"os"
 
 	"github.com/spf13/cobra"
+
+	"github.com/giantswarm/formica/task"
 )
 
 var (
@@ -23,10 +25,40 @@ func startRun(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	err = newController.Start(req)
+	action := func() error {
+		err = newController.Start(req)
+		if err != nil {
+			return maskAny(err)
+		}
+
+		return nil
+	}
+
+	taskObject, err := newTask.Create(action)
 	if err != nil {
 		fmt.Printf("%#v\n", maskAny(err))
 		os.Exit(1)
+	}
+
+	if !globalFlags.NoBlock {
+		taskObject, err = newTask.WaitForFinalStatus(taskObject, nil)
+		if err != nil {
+			fmt.Printf("%#v\n", maskAny(err))
+			os.Exit(1)
+		}
+
+		if task.HasFailedStatus(taskObject) {
+			if err != nil {
+				if req.SliceIDs == nil {
+					fmt.Printf("Failed to start group '%s'\n", req.Group)
+				} else if len(req.SliceIDs) == 0 {
+					fmt.Printf("Failed to start all slices of group '%s'\n", req.Group)
+				} else {
+					fmt.Printf("Failed to start %d slices for group '%s': %v", len(req.SliceIDs), req.Group, req.SliceIDs)
+				}
+				os.Exit(1)
+			}
+		}
 	}
 
 	if req.SliceIDs == nil {
--- a/cli/start.go
+++ b/cli/start.go
@@ -33,7 +33,7 @@ func startRun(cmd *cobra.Command, args []string) {
 		Request:    req,
 		Descriptor: "start",
 		NoBlock:    globalFlags.NoBlock,
-		TaskObject: taskObject,
+		TaskID:     taskObject.ID,
 		Closer:     nil,
 	})
 }
--- a/cli/start.go
+++ b/cli/start.go
@@ -5,8 +5,6 @@ import (
 	"os"
 
 	"github.com/spf13/cobra"
-
-	"github.com/giantswarm/formica/task"
 )
 
 var (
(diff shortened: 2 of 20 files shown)


task/backend.go
```diff
@@ -0,0 +1,6 @@
+package task
+
+type Backend interface {
+	Get(string) (*TaskObject, error)
+	Set(*TaskObject) error
```
xh3b4sd: Here I need to put the argument names.

xh3b4sd:
Oh yeah, note that this is WIP and for now to make sure we are going into the right direction. Please no comments about missing tests and missing comments and stuff. This will follow.

cli/formicactl.go
```diff
@@ -49,12 +52,16 @@ var (
 			newController = controller.NewController(newControllerConfig)
 
 			newFileSystem = filesystemreal.NewFileSystem()
+
+			newTaskConfig := task.DefaultTaskConfig()
+			newTask = task.NewTask(newTaskConfig)
```
zeisss: why does the CLI initialize a new task?

cli/formicactl.go
```diff
@@ -49,12 +52,16 @@ var (
...
 		},
 	}
 )
 
 func init() {
 	MainCmd.PersistentFlags().StringVar(&globalFlags.FleetEndpoint, "fleet-endpoint", "unix:///var/run/fleet.sock", "endpoint used to connect to fleet")
+	MainCmd.PersistentFlags().BoolVar(&globalFlags.NoBlock, "no-block", false, "block on syncronous actions or not")
```
zeisss: s/syncronous/synchronous/

cli/formicactl.go
xh3b4sd: I suggest having a general look at the PR and then asking questions. Maybe naming needs to be fixed here in some places. A "task" here refers to the task managing functionality, not a task per se.

zeisss:
ok, general questions:
- The concept seems to be to have a task lib for the CLI to use to do asynchronous calls to the controller. so the CLI commands need to explicitly use this. In case we want to add middlewares for a http server, we would need to replicate this logic into the middlewares. Q: Wouldn't it make sense to integrate this into the controller itself so the cli and middlewares are not concerned with the tasks itself?

xh3b4sd:
Please provide an idea about how the interface of the controller then needs to look like, because this "no-block" information needs to be provided.

zeisss:
Well, from the CLI view like this:

```
task, err := controller.Start(req)
if err != nil { ... } 

if !noBlock {
  latestTaskState, err := controller.WaitForTask(task.Id)
  if err != nil {
     fmt.Println("Your task failed!...")
  } else {
    fmt.Println("All slices are up and running")
  }
}
```

xh3b4sd:
Now I got your idea. This makes sense. Lets do that. Anything else?

zeisss:
I would make the task package a subpackage of the controller. Its internal stuff, no need to add more packages on the first level.

task/task.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	go func() {
+		err := action()
+		if err != nil {
+			fmt.Printf("[E] Task.Action failed: %#v\n", maskAny(err))
+			return
+		}
+
+		_, err = t.MarkAsFailed(taskObject)
```
zeisss: this logic in the go fun makes no sense. I think we are missing an `if` here

task/task.go
```diff
@@ -0,0 +1,151 @@
...
+type Action func() error
+
+type TaskObject struct {
+	ID string
+
+	ActiveStatus ActiveStatus
+	FinalStatus  FinalStatus
+}
+
+type Task interface {
```
zeisss: How about calling this `Service` ? Thus it's `task.Service` and would give a rather clear idea of what it is about. the controller could call it `TaskService` or `Tasks`

task/status.go
```diff
@@ -0,0 +1,39 @@
+package task
+
+type ActiveStatus string
+
+const (
+	StatusStarted ActiveStatus = "started"
```
zeisss, line 7: How about another pending state?

zeisss:
done

task/status.go
xh3b4sd, line 7: I see it as "started" implies pending. What should be the benefit?

task/task.go
xh3b4sd: No strong feelings. I can do that.

task/task.go
xh3b4sd: Right. This needs to be within the former condition in case the action fails.

task/status.go
zeisss, line 7: pending for me means "not started yet", so the order would be `pending -> started -> stopped`. But it's not necessary. Lets see how far we get with this :)

xh3b4sd, line 7: +1

xh3b4sd:
RFR @zeisss @JosephSalisbury I would like to merge this and create a separate PR only for testing afterwards. The PR is getting to big and since the master is not critical to provide a certain coverage I am willing to ignore the coverage state for a day.

cli/start.go
```diff
@@ -23,17 +23,17 @@ func startRun(cmd *cobra.Command, args []string) {
...
-		fmt.Printf("Started group '%s'\n", req.Group)
-	} else if len(req.SliceIDs) == 0 {
-		fmt.Printf("Started all slices of group '%s'\n", req.Group)
-	} else {
-		fmt.Printf("Started %d slices for group '%s': %v", len(req.SliceIDs), req.Group, req.SliceIDs)
-	}
+	maybeBlockWithFeedback(blockWithFeedbackCtx{
+		Request:    req,
+		Descriptor: "start",
+		NoBlock:    globalFlags.NoBlock,
```
zeisss, line 35: I like that the function is not relying on the global flag and instead we pass it as an argument. :cake:

controller/controller.go
```diff
@@ -206,6 +253,11 @@ func (c controller) GetStatus(req Request) ([]fleet.UnitStatus, error) {
 	return status, maskAny(err)
 }
 
+func (c controller) WaitForTask(taskObject *task.TaskObject, closer <-chan struct{}) (*task.TaskObject, error) {
```
zeisss: do we need the whole task object or is a `taskID` sufficient? That would make a nicer interface in my opinion

task/backend.go
```diff
@@ -0,0 +1,10 @@
+package task
+
+// Backend represents some storage solution to persist task objects.
+type Backend interface {
```
zeisss: I would suggest naming this `Storage` or `Repository` then. `Backend` is a bit too generic IMO.

controller/controller.go
```diff
@@ -38,23 +44,28 @@ func DefaultConfig() Config {
 type Controller interface {
 	// Submit schedules a group on the configured fleet cluster. This is done by
 	// setting the state of the units in the group to loaded.
-	Submit(req Request) error
+	Submit(req Request) (*task.TaskObject, error)
```
zeisss, line 47: Do we want to define an `interface` in the `controller/` package, so the client doesn't need knowledge of the other packages?

task/task_service.go
```diff
@@ -0,0 +1,190 @@
...
+
+	err := ts.PersistState(taskObject)
+	if err != nil {
+		return nil, maskAny(err)
+	}
+
+	return taskObject, nil
+}
+
+func (ts *taskService) PersistState(taskObject *TaskObject) error {
```
zeisss, line 159: does this need to be public?

task/task_service.go
```diff
@@ -0,0 +1,190 @@
...
+		ActiveStatus: StatusStarted,
+		FinalStatus:  "",
+	}
+
+	go func() {
+		err := action()
+		if err != nil {
+			_, markErr := ts.MarkAsFailedWithError(taskObject, err)
+			if markErr != nil {
+				fmt.Printf("[E] Task.MarkAsFailed failed: %#v\n", maskAny(markErr))
```
zeisss, line 102: do we want a logger?

controller/controller.go
xh3b4sd: That would mean to change the interface within the controller and the task service. I did it as it is on purpose. So all the time you deal with the task object. IMO this is already nice.

zeisss: ~~> and the task service~~

~~Why? The taskservice returns a struct which also fulfills the interface the controller defines and returns.~~

controller/controller.go
xh3b4sd, line 47: No strong feelings. I don't see that much benefit. Looking at fleet this is the same, and there, it is even worse. See https://github.com/giantswarm/app-service/blob/master/service/scheduler-service/fleet_client_scheduler.go#L12-L17

task/task_service.go
xh3b4sd, line 159: No. My intention was to define an interface that defines the interface of a implementation. So the interface provides all building blocks to make the service work. The interface here does not care about visibility, but logic. Persisting state is a logical component that needs to be fulfilled. Thus the interface requires that logic to be implemented.

task/task_service.go
xh3b4sd, line 102: Yes we do want a logger. Would you mind adding a card to huboard? You are used to that management layer stuff.

task/task_service.go
```diff
@@ -0,0 +1,190 @@
...
+		return maskAny(err)
+	}
+
+	return nil
+}
+
+func (ts *taskService) WaitForFinalStatus(taskObject *TaskObject, closer <-chan struct{}) (*TaskObject, error) {
+	for {
+		select {
+		case <-closer:
```
zeisss, line 174: :cake:

task/task_service.go
zeisss, line 159: :white_check_mark:

task/task_service.go
```diff
@@ -0,0 +1,190 @@
...
+
+	err := ts.PersistState(taskObject)
+	if err != nil {
+		return nil, maskAny(err)
+	}
+
+	return taskObject, nil
+}
+
+func (ts *taskService) FetchState(taskObject *TaskObject) (*TaskObject, error) {
```
xh3b4sd: Lets simply put the task ID here.

task/task_service.go
```diff
@@ -0,0 +1,190 @@
...
+func (ts *taskService) PersistState(taskObject *TaskObject) error {
+	err := ts.Backend.Set(taskObject)
+	if err != nil {
+		return maskAny(err)
+	}
+
+	return nil
+}
+
+func (ts *taskService) WaitForFinalStatus(taskObject *TaskObject, closer <-chan struct{}) (*TaskObject, error) {
```
xh3b4sd: Lets simply put the task ID here.

task/task_service.go
zeisss, line 102: here ya go: https://github.com/giantswarm/giantswarm/issues/536

controller/controller.go
zeisss: Discussed :white_check_mark:

zeisss:
LGTM, although the backend thing would be lovely, but low prio

task/backend.go
xh3b4sd: Storage it is.

controller/controller.go
xh3b4sd: :cake:

PR #35, opened 2016-02-25
added tests for task package

xh3b4sd:


task/task_service_test.go
```diff
@@ -0,0 +1,155 @@
...
+	action := func() error {
+		fmt.Printf("%#v\n", 2)
+
+		time.Sleep(1 * time.Second)
+		fmt.Printf("%#v\n", 7)
+
+		return nil
+	}
+
+	fmt.Printf("%#v\n", 1)
```
xh3b4sd: @zeisss @JosephSalisbury please have a look. The weird thing is that this test does work when having these comments in there. When I remove the `fmt.Printf` calls, the test fails. Do you have any idea?

xh3b4sd:
RFR @zeisss @JosephSalisbury

task/task_service_test.go
zeisss: race conditions! (I will check :)

task/task_service.go
```diff
@@ -25,6 +22,9 @@ type TaskObject struct {
 	// FinalStatus represents any status that is final. A task having this status
 	// will not change its status anymore.
 	FinalStatus FinalStatus
+
+	// ID represents the task identifier.
+	ID string
```
zeisss, line 27: do we order by ABC now?

task/task_service_test.go
```diff
@@ -0,0 +1,155 @@
+package task
+
+import (
+	"fmt"
+	"testing"
+	"time"
+)
+
+func Test_Task_TastService_Create_Success(t *testing.T) {
```
zeisss: TastService?

task/task_service_test.go
zeisss: Ok, I think I have it. We are using pointers to the Task everywhere, even in the MemoryStorage. This means the execution as well as the WaitForFinalStatus() all read and write on the same object. 

If you leave this line and run the test with `-v` you will get the same effect. I guess its not the print itself but just any statement triggers the racecondition more evenly.

xh3b4sd: > If you leave this line and run the test with -v you will get the same effect.

I don't understand this. 

How to fix the test?

zeisss: The other thing is as following: `WaitForFinalStatus` always sleeps 1 sec between checks. Even if you send the abort signal after 200ms, it will sleep 1s and then see the abort. This gives the other goroutine enough time to actually finish and update the task in the storage.

zeisss: I have a test that works, shall I commit it?

task/task_service.go
xh3b4sd, line 27: I did that yes.

task/task_service_test.go
xh3b4sd: Ah good catch. This was the problem. I made the wait configurable.

task/task_service_test.go
xh3b4sd: Fixed

task/task_service.go
```diff
@@ -173,7 +177,7 @@ func (ts *taskService) WaitForFinalStatus(taskID string, closer <-chan struct{})
 		select {
 		case <-closer:
 			return nil, nil
-		default:
+		case <-time.After(1 * time.Second):
```
zeisss: oops, will fix to use `ts. WaitSleep`

xh3b4sd:
I would like to merge this. Who is up for a LGTM?

zeisss:
LGTM

PR #36, opened 2016-02-25
Fixes racecondition in tests

zeisss:
a fix for the race condition issue.

### DIFF
--- a/task/memory_storage.go
+++ b/task/memory_storage.go
@@ -9,7 +9,7 @@ import (
 func NewMemoryStorage() Storage {
 	newStorage := &memoryStorage{
 		Mutex:   sync.Mutex{},
-		Storage: map[string]*TaskObject{},
+		Storage: map[string]TaskObject{},
 	}
 
 	return newStorage
@@ -17,7 +17,7 @@ func NewMemoryStorage() Storage {
 
 type memoryStorage struct {
 	Mutex   sync.Mutex
-	Storage map[string]*TaskObject
+	Storage map[string]TaskObject
 }
 
 func (mb *memoryStorage) Get(taskID string) (*TaskObject, error) {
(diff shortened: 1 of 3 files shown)


task/memory_storage.go
```diff
@@ -9,7 +9,7 @@ import (
 func NewMemoryStorage() Storage {
 	newStorage := &memoryStorage{
 		Mutex:   sync.Mutex{},
-		Storage: map[string]*TaskObject{},
```
zeisss, line 12: this ensures we are creating copies of our objects - so modifications to the object after saving does not change it anymore. same as a stored task in redis or etcd would be

task/task_service.go
```diff
@@ -173,7 +173,7 @@ func (ts *taskService) WaitForFinalStatus(taskID string, closer <-chan struct{})
 		select {
 		case <-closer:
 			return nil, nil
-		default:
```
zeisss, line 180: this fixes the fact that `WaitForFinalStatus` always blocked 1s which lead to the race condition with in the Create test

task/task_service_test.go
```diff
@@ -107,46 +107,37 @@ func Test_Task_TastService_Create_Wait(t *testing.T) {
...
-		fmt.Printf("%#v\n", 2)
-
+		// Just something to do, so the task blocks
 		time.Sleep(1 * time.Second)
-		fmt.Printf("%#v\n", 7)
 
 		return nil
 	}
 
-	fmt.Printf("%#v\n", 1)
```
zeisss: some renaming I did while playing around here. Not really needed, but might make it more readable

zeisss:
ping @xh3b4sd

task/task_service.go
```diff
@@ -173,7 +173,7 @@ func (ts *taskService) WaitForFinalStatus(taskID string, closer <-chan struct{})
...
+		case <-time.After(1 * time.Second):
```
xh3b4sd, line 180: Please merge the origin feature branch and use the configured `WaitSleep` here.

xh3b4sd:
Please fix the comment. LGTM

PR #37, opened 2016-02-25
Update question

zeisss:
https://github.com/giantswarm/giantswarm/issues/520

Came up in planning. Here for some discussion.

### DIFF
--- a/docs/update-design-question.md
+++ b/docs/update-design-question.md
@@ -1,0 +1,45 @@
+Guys, we have a question for Formica where everyone is invited to participate deciding.
+
+So, formica currently is a tool for managing groups of units. Example:
+
+```
+$ formicactl submit elasticsearch@{1,2}
+$ formciactl start elasticsearch@{1,2}
+$ formicactl status elasticsearch
+Group           Unit  <more headers>
+elasticsearch@1 *     active  active  running 10.0.4.100
+elasticsearch@2 *     active  active  running 10.0.4.101
+```
+
+As you see, it is very similar to `fleetctl`. Next we want to implement an update
+functionality, similar (but better ;) than what `releaseit` currently does.
+
+## Question 1
+In the example above we have used `1` and `2` for the slices. When updating, we
+want the users to have the option to create new instances first. This raises where
+the names for those slices come from.
+
+* Option 1: Just increment the found ids. So if we find `7,8,9` we continue with
+  `10, 11, ..`
+* Option 2: Use random IDs. This would imply changes to the idea that user dictates
+  these names as shown above.
+* Option 3: Extend the slice with a timestamp so we can reuse the old slice but still differentiate
+  them from existing slices.
+
+## Question 2
+
+The second question is how the update command it self looks like.
+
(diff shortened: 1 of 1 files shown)


docs/update-design-question.md
```diff
@@ -0,0 +1,45 @@
...
+* Option 2: Use random IDs. This would imply changes to the idea that user dictates
+  these names as shown above.
+* Option 3: Extend the slice with a timestamp so we can reuse the old slice but still differentiate
+  them from existing slices.
+
+## Question 2
+
+The second question is how the update command it self looks like.
+
+1. `$ formicactl update --sleep-time=1m --max-unavailable=50% --max-surge=100% elasticsearch`
```
puja108, line 33: might be obvious but what does sleep-time do? why is it needed?

docs/update-design-question.md
```diff
@@ -0,0 +1,45 @@
...
+
+As you see, it is very similar to `fleetctl`. Next we want to implement an update
+functionality, similar (but better ;) than what `releaseit` currently does.
+
+## Question 1
+In the example above we have used `1` and `2` for the slices. When updating, we
+want the users to have the option to create new instances first. This raises where
+the names for those slices come from.
+
+* Option 1: Just increment the found ids. So if we find `7,8,9` we continue with
```
hectorj2f, line 22: Option1: I would find it a bit awkard whenever I start to remove/add units.. Moreover I'd like the units to be shown by age (older/newer).

Option2: Random ids is the way to go in my opinion. In fact, the group should be elasticsearch(pod) and its ID elasticsearch@123131. If you wanna stop a pod-group you just use group names plus the ID. elasticsearch@RANDOM_ID. I know that this differs with what we have but it'd be closer to the way this kind of things are managed by other tools, common cli's, or kubelet, rkt,

Option 3: It isn't clear to me, so you keep the same Group 'elasticsearhc@2' but with a different timestamp. I don't like this option

docs/update-design-question.md
hectorj2f, line 33: Vote to 1. To follow a similar standard

zeisss, line 33: sleep time in our case is to give our ambassadors a time actually work. Kubernetes has sthg similar, where they e.g. wait for 1min to and verify the new container actually stayed up in that time (and didn't die again).

docs/update-design-question.md
teemow, line 22: This part is an identifier for the instance. Incrementing is probably not a good idea. Just think of two people working with the system at the same time. If you think of higher level orchestration it would be helpful to receive that instance identifier (eg. our instance id or similar). 

I took a quick look at how `systemd-run` handles this. This is not a perfect match but you can define a `--unit` or it creates a random one.

docs/update-design-question.md
puja108, line 33: ok, got it, is it also called sleep-time in k8s? sounds not very clear as to what is sleeping and what to use it for, but ok if that's a "known"

zeisss:
FYI: After some more discussions and involving @marians we will now head for an CLI similar to kubernetes (which also allows the other way) and using random, but very short IDs.

PR #39, opened 2016-02-26
Add golint to CI

JosephSalisbury:
Related to https://github.com/giantswarm/formica/issues/38

This adds `golint` to the build process, meaning that we can inspect its output as part of the PR process. Note - `golint` will not fail the build if it finds anything.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,9 +65,8 @@ test: $(SOURCE) VERSION .gobuild
 	    $(TEST_COMMAND)
 
 lint:
-	for source_file in $(SOURCE); do \
-		go vet -x $$source_file ; \
-	done
+	go vet -x ./...
+	golint ./...
 
 ci-build: $(SOURCE) VERSION .gobuild
 	echo Building for $(GOOS)/$(GOARCH)
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ before_install:
 - wget https://downloads.giantswarm.io/builder/0.11.0/builder
 - chmod +x ./builder
 - export PATH=$PATH:$PWD
+- go get -u github.com/golang/lint/golint
 
 script:
 - make lint


Makefile
```diff
@@ -65,9 +65,8 @@ test: $(SOURCE) VERSION .gobuild
 	    $(TEST_COMMAND)
 
 lint:
-	for source_file in $(SOURCE); do \
-		go vet -x $$source_file ; \
-	done
```
JosephSalisbury, line 70: This was a dumb way to do this, so let's do it better! :D

JosephSalisbury:
@giantswarm/cena RFR

Makefile
zeisss, line 70: more tools that need to be installed locally. Are they not part of the go docker image?

xh3b4sd:
Looking at the CI report shows it does not work by working. Golint complains but exits with 0.

JosephSalisbury:
@xh3b4sd It's too false-positivey to fail the build. It explicitly says not to use it to fail builds in their [README](https://github.com/golang/lint#purpose) (like I did in the PR description :trollface:)

Makefile
JosephSalisbury, line 70: `golint` is not part of the golang docker image, sadly. With regards to linting, my reckoning was to make sure it was part of the CI system, over running it locally.

xh3b4sd:
So what is the point of integrating it here in the CI then? Are you really going to check the builds before merging? To be honest, this feels error prone.

JosephSalisbury:
I tend to check the build page, and I like having CI check these things - a little extra information isn't a bad thing.

xh3b4sd:
I tend to be sceptic here. No strong feelings. Go ahead.

JosephSalisbury:
Sceptic is good <3

PR #43, opened 2016-02-29
*: Fixed various go vet and golint warnings

### LINKED ISSUE #38: Golint Stutter Warnings
My editor warns me about our current names:

> task/task_service.go:14:6: type name will be used as task.TaskObject by other packages, and that stutters; consider calling this Object
> task/task_service.go:32:6: type name will be used as task.TaskService by other packages, and that stutters; consider calling this Service
> task/task_service.go:63:6: type name will be used as task.TaskServiceConfig by other packages, and that stutters; consider calling this ServiceConfig

I think I agree here. `TaskObject` could simply be `Task`, `TaskService` become `Service` etc.

Opinions @giantswarm/cena ?

zeisss:
Fixes #38. These are only renames done by `gofmt` and a few manualy reverts, e.g. to keep `TaskService` inside the `Controller` for the field name.

### DIFF
--- a/task/task_service.go
+++ b/task/task_service.go
@@ -10,8 +10,8 @@ import (
 // Action represents any work to be done when executing a task.
 type Action func() error
 
-// TaskObject represents a task that is executable.
-type TaskObject struct {
+// Task represents a task that is executable.
+type Task struct {
 	// ActiveStatus represents a status indicating activation or deactivation.
 	ActiveStatus ActiveStatus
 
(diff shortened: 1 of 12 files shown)


task/task_service.go
```diff
@@ -10,67 +10,50 @@ import (
 // Action represents any work to be done when executing a task.
 type Action func() error
 
-// TaskObject represents a task that is executable.
-type TaskObject struct {
```
zeisss, line 14: moved to `task.go`

zeisss:
ping @giantswarm/cena

task/task_service.go
xh3b4sd, line 14: Why?

xh3b4sd:
I think some changes do not make sense but I don't have strong feelings here. Go ahead.

task/task_service.go
zeisss, line 14: the file was there and it felt logic for me to put the type `Task` into the file `task.go`. Since you ask, I will revert. there is no need for this change by itself.

PR #48, opened 2016-03-02
test controller statuses

xh3b4sd:
Epic: https://github.com/giantswarm/giantswarm/issues/522

This is simply to add more tests.

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -313,17 +313,7 @@ func Test_matchesGroupSlices(t *testing.T) {
 // with a mock.
 func givenController() (Controller, *fleetMock) {
 	newFleetMockConfig := defaultFleetMockConfig()
-	newFleetMock := newFleetMock(newFleetMockConfig)
-
-	newTaskServiceConfig := task.DefaultConfig()
-	newTaskService := task.NewTaskService(newTaskServiceConfig)
-
-	newControllerConfig := DefaultConfig()
-	newControllerConfig.Fleet = newFleetMock
-	newControllerConfig.TaskService = newTaskService
-	newControllerConfig.WaitCount = 1
-	newControllerConfig.WaitSleep = 10 * time.Millisecond
-	newController := NewController(newControllerConfig)
+	newController, newFleetMock := givenControllerWithConfig(newFleetMockConfig)
 
 	return newController, newFleetMock
 }
@@ -531,7 +521,9 @@ func TestController_Status_ErrorOnMismatchingSliceIDs(t *testing.T) {
 	mock.AssertExpectationsForObjects(t, fleetMock.Mock)
 }
 
-func TestController_WaitForStatus_001(t *testing.T) {
+// TestController_WaitForStatus_Success tests the normal behaviour of
+// Controller.WaitForStatus to ensure it works as expected.
+func TestController_WaitForStatus_Success(t *testing.T) {
 	RegisterTestingT(t)
 
 	// Mocks
@@ -541,13 +533,13 @@ func TestController_WaitForStatus_001(t *testing.T) {
 	newFleetMockConfig.FirstCustomMockStatus = []fleet.UnitStatus{
 		{
 			Current: "loaded",
-			Desired: "*",
+			Desired: "loaded",
 			Machine: []fleet.MachineStatus{
 				{
 					ID:            "test-id",
 					IP:            net.ParseIP("10.0.0.101"),
 					SystemdActive: "activating",
-					SystemdSub:    "*",
+					SystemdSub:    "start-pre",
 					UnitHash:      "test-hash",
 				},
 			},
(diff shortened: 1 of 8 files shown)


xh3b4sd:
RFR @zeisss @JosephSalisbury @Nesurion

controller/controller_test.go
```diff
@@ -310,16 +312,38 @@ func Test_matchesGroupSlices(t *testing.T) {
 // givenController returns a controller where the fleet backend is replaced
 // with a mock.
 func givenController() (Controller, *fleetMock) {
-	fleetMock := fleetMock{}
+	newFleetMockConfig := defaultFleetMockConfig()
```
zeisss, line 315: do we just want to call `givenControllerWithConfig` here?

controller/controller_test.go
```diff
@@ -470,3 +530,166 @@ func TestController_Status_ErrorOnMismatchingSliceIDs(t *testing.T) {
 	Expect(status).To(BeEmpty())
 	mock.AssertExpectationsForObjects(t, fleetMock.Mock)
 }
+
+func TestController_WaitForStatus_001(t *testing.T) {
```
zeisss: I really would like to see a descriptive name of this test instead of `001`

controller/fleet_mock_test.go
```diff
@@ -37,6 +67,17 @@ func (fm *fleetMock) GetStatusWithExpression(exp *regexp.Regexp) ([]fleet.UnitSt
...
 func (fm *fleetMock) GetStatusWithMatcher(f func(string) bool) ([]fleet.UnitStatus, error) {
-	args := fm.Called(f)
-	return args.Get(0).([]fleet.UnitStatus), args.Error(1)
+	if fm.UseTestifyMock {
+		args := fm.Called(f)
+		return args.Get(0).([]fleet.UnitStatus), args.Error(1)
+	} else if fm.UseCustomMock {
+		fm.CustomMockUsed++
+		if fm.CustomMockUsed <= 3 {
+			return fm.FirstCustomMockStatus, nil
```
zeisss, line 76: you should be able to do this with testify, no need for special logic.

```
mock.On("GetS..."..).Return(FirstCustomMockStatus).Times(3)
mock.On("Get"..).Return(LastCustomMockStatus)
```

xh3b4sd, line 76: I don't understand what you mean with `"GetS..."..` and `"Get"..`.

controller/controller_test.go
xh3b4sd: What name would you like to see?

controller/controller_test.go
xh3b4sd, line 315: Why should that be better?

zeisss, line 315: because the rest of the function here is an exact 1:1 copy of `givenControllerWithConfig()`

controller/controller_test.go
zeisss: Something that gives me a hint what this test wants to prove. At first glance it seems to simulate a unit switching from activating to running, so maybe `TestController_WaitForStatus_UnitActivatingToRunning` ?

controller/controller_test.go
```diff
@@ -470,3 +530,166 @@ func TestController_Status_ErrorOnMismatchingSliceIDs(t *testing.T) {
...
+	newFleetMockConfig.FirstCustomMockStatus = []fleet.UnitStatus{
+		{
+			Current: "loaded",
+			Desired: "*",
+			Machine: []fleet.MachineStatus{
+				{
+					ID:            "test-id",
+					IP:            net.ParseIP("10.0.0.101"),
+					SystemdActive: "activating",
+					SystemdSub:    "*",
```
zeisss: can we put a real state here? e.g. `pre-start`

controller/fleet_mock_test.go
zeisss, line 76: the method name and arguments that you expect to get called.

controller/controller_test.go
xh3b4sd: `I_Don't_Like_These_Kind_Of_Test_Names_Because_Even_Making_It_Longer_Questions_How_Good_The_Hint_Really_Is`. What about simply documenting the necessary information with some comments? Then you have a nice description, a proper hint, and the method name stays nice and at least kind of short.

controller/controller_test.go
xh3b4sd, line 315: Now I got you. Sure. Done.

controller/controller_test.go
xh3b4sd: Done

controller/controller_test.go
xh3b4sd: Changed the names and added comments.

controller/fleet_mock_test.go
xh3b4sd, line 76: I cannot make it work that way. In case you want to invest time here lets have a look at it together. Otherwise I vote for keeping it as it is right now.

zeisss:
Lets check the mock stuff together, until then: LGTM

PR #51, opened 2016-03-03
Add example

Nesurion:
adds example unit files


Nesurion:
misplaced commit. removing...

PR #52, opened 2016-03-03
document basic concepts

xh3b4sd:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -4,7 +4,6 @@ Here we document important concepts and functionality Inago supports.
 ### table of contents
 
 - [What is the goal of Inago?](goal.md)
-- [What do we expect from the users?](expect.md)
+- [What unit file structure is required?](structure.md)
 - [What are the terms we use and what do they mean?](terms.md)
-- [What is the structure in fleet we generate?](fleet_structure.md)
 - [What is slice expansion?](slice_expansion.md)
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,5 +1,10 @@
 # docs
-Here we document important concepts and functionality formica supports.
+Here we document important concepts and functionality Inago supports.
 
 ### table of contents
-- [slice expansion](slice_expansion.md)
+
+- [What is the goal of Inago?](goal.md)
+- [What do we expect from the users?](expect.md)
+- [What are the terms we use and what do they mean?](terms.md)
+- [What is the structure in fleet we generate?](fleet_structure.md)
+- [What is slice expansion?](slice_expansion.md)
--- a/docs/goal.md
+++ b/docs/goal.md
@@ -2,6 +2,4 @@
 Inago is a deployment tool that manages groups of unit files to deploy them to
 a fleet cluster similar to `fleetctl`. Since `fleetctl` is quite limited, Inago
 aims to abstract units away and provide more sugar on top like update
-strategies. That way the user can manage unit files more easily. All actions one
-can apply to a unit using `fleetctl`, can also be applied towards a group using
-`inagoctl`.
+strategies. That way the user can manage unit files more easily.
--- a/docs/goal.md
+++ b/docs/goal.md
@@ -1,0 +1,7 @@
+# What is the goal of Inago?
+Inago is a deployment tool that manages groups of unit files to deploy them to
+a fleet cluster similar to `fleetctl`. Since `fleetctl` is quite limited, Inago
+aims to abstract units away and provide more sugar on top like update
+strategies. That way the user can manage unit files more easily. All actions one
+can apply to a unit using `fleetctl`, can also be applied towards a group using
+`inagoctl`.
--- a/docs/terms.md
+++ b/docs/terms.md
@@ -1,0 +1,26 @@
+# What are the terms we use and what do they mean?
+
+### Group
+A group is a collection of unit files tied together. This can be seen as
+equivalent to kubernetes pods. Units within a Inago group are scheduled
+together on the same host.
+
+Since a group represents many units, Inago aims to abstract units away, so
+the user can manage a group more easily. Having this given, a group can be
+treated like a single unit. All actions one can apply to a unit using
+`fleetctl`, can also be applied towards a group using `inagoctl`.
+
+A group is named after the directory its unit files live in. Having two units
+`mygroup-foo@.service` and `mygroup-bar@.timer` in the directory `mygroup`,
+leads to the Inago group `mygroup`. So creating a directory and putting units
+into it, creates a new group as Ingao is able to understand it.
+
+### Slice
+A slice is a scalable group instance. The kubernetes equivalent would be a
+replica. Slices are a scaled version of a group. So you can scale up your group
+using slice expansion. See also [What is slice expansion?](slice_expansion.md).
+
+### Instance
+A group instance is a group in a fleet cluster. Having a group definition on
+your local file system and deploying this using Inago to a fleet cluster,
+creates a group instance. This group instance may be one of multiple slices.
--- a/docs/terms.md
+++ b/docs/terms.md
@@ -2,8 +2,8 @@
 
 ### Group
 A group is a collection of unit files tied together. This can be seen as
-equivalent to kubernetes pods. Units within a Inago group are scheduled
-together on the same host.
+equivalent to kubernetes pods. Units within an Inago group are scheduled
+together on the same host by convention.
 
 Since a group represents many units, Inago aims to abstract units away, so
 the user can manage a group more easily. Having this given, a group can be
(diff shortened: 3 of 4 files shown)


docs/README.md
```diff
@@ -1,5 +1,10 @@
 # docs
-Here we document important concepts and functionality formica supports.
+Here we document important concepts and functionality Inago supports.
 
 ### table of contents
-- [slice expansion](slice_expansion.md)
+
+- [What is the goal of Inago?](goal.md)
+- [What do we expect from the users?](expect.md)
```
xh3b4sd: I copied this from huboard but I don't know what that means.

docs/README.md
```diff
@@ -1,5 +1,10 @@
...
+- [What are the terms we use and what do they mean?](terms.md)
+- [What is the structure in fleet we generate?](fleet_structure.md)
```
xh3b4sd: Same here. What is the idea behind this?

xh3b4sd:
RFC @zeisss @JosephSalisbury @Nesurion 

What do you think? What to add? Do we want to have examples?

docs/terms.md
```diff
@@ -0,0 +1,26 @@
+# What are the terms we use and what do they mean?
+
+### Group
+A group is a collection of unit files tied together. This can be seen as
+equivalent to kubernetes pods. Units within a Inago group are scheduled
```
JosephSalisbury: an Inago group

docs/goal.md
```diff
@@ -0,0 +1,7 @@
+# What is the goal of Inago?
+Inago is a deployment tool that manages groups of unit files to deploy them to
+a fleet cluster similar to `fleetctl`. Since `fleetctl` is quite limited, Inago
+aims to abstract units away and provide more sugar on top like update
+strategies. That way the user can manage unit files more easily. All actions one
+can apply to a unit using `fleetctl`, can also be applied towards a group using
```
zeisss: > All actions one can apply to a unit using `fleetctl`, can also be applied towards a group using `inagoctl`.

Thats not really true. We have opted to not support `load/unload`, `status` behaves differently (its more like the `list-*`) and `cat` is still to be debated.

docs/terms.md
zeisss: The thing is, Inago does not enforce the "scheduled on the same host". It's one of the assumptions we do. Not sure how to incorporate this here though.

docs/terms.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Slice
+A slice is a scalable group instance. The kubernetes equivalent would be a
+replica. Slices are a scaled version of a group. So you can scale up your group
+using slice expansion. See also [What is slice expansion?](slice_expansion.md).
+
+### Instance
+A group instance is a group in a fleet cluster. Having a group definition on
+your local file system and deploying this using Inago to a fleet cluster,
+creates a group instance. This group instance may be one of multiple slices.
```
zeisss, line 26: > This group instance may be one of multiple slices.

Uhm, so far we have talked about a group either having one instance (the no `@` case) or multiple slices (the `@` case).

zeisss:
Cool! Can we add an assumptions.md file which lists our assumptions? We don't need to make text out of it yet, just that it's there.

xh3b4sd:
@zeisss maybe you reply to my comment. I was asking what this assumptions are or what this is about. See https://github.com/giantswarm/formica/pull/52#discussion-diff-54929779 and others.

docs/terms.md
xh3b4sd, line 26: It is hard for me to get the term `instance` into my mental model. For me there are only slices. One or many slices form a group. Should I simply remove the last sentence?

docs/terms.md
xh3b4sd: Even the tool does not enforce it, this should be the convention, shouldn't it?

docs/README.md
zeisss: we expect them to provide us folders with unit files in a certain format.

docs/README.md
zeisss: Documenting how the model that Inago provides maps to the unit files we actually schedule in fleet.

docs/terms.md
zeisss, line 26: > It is hard for me to get the term instance into my mental model

Then we need to discuss this better or find a different term. Sorry!

> For me there are only slices.

Whats about groups where you don't want slices? No `@` sign, e.g. our unit replicators, the clean services/timers etc.

> One or many slices form a group.

Mh, I would rather say that the Group is like the definition and it can have zero to `n` slices.

zeisss, line 26: > Should I simply remove the last sentence?

Not sure, lets cleanup our terms first.

docs/terms.md
zeisss: yes, that is the assumption we do on what the user provides us. Might just be a small phrasing issue, as this felt on first read like sthg that is enforced by Inago.

docs/README.md
xh3b4sd: I am not sure what structure this means then. In fleet there are simply units. You define a group based on the unit files within the group directory and then you have this units scheduled to the fleet cluster. I don't know what to document here. Maybe you want to simply write this on your own. Feel free to hijack the PR.

docs/README.md
xh3b4sd: I don't understand this. Unit file content doesn't need to be in a certain format. The only requirement is the unit file names. 

Maybe we have a communication issue here. When I read `What do we expect from the users?`, I think `Don't be dumb`, but this is probably not what you mean. May you want to write this part of the documentation yourself like I already stated in my other comment about the fleet structure (which I also do not understand properly).

docs/terms.md
xh3b4sd: Will add the conventional part.

docs/goal.md
xh3b4sd: Will remove this part.

docs/terms.md
xh3b4sd, line 26: IMO you always have a slice. No matter if there is a `@` or not. Once instance of a group is a slice for me. So I wonder why we need the term `instance` when we have `slice`. Why making a difference in wording only because the unit name changes? At the end you have something running on a fleet cluster that was defined by a scheme. The scheme, the template, that is the group on your local file system, is a blueprint to schedule something. This something is either a slice of an instance. Please lets keep it simple and use one word. Since we heavily used the term `slice` in the code, I vote to simply throw `instance` away.

docs/README.md
zeisss: then lets drop it for now. I just made a few notes in the issues - thats not a hard written list to document.

docs/README.md
zeisss: > Unit file content doesn't need to be in a certain format. 

Yes, they need to be. There must be a folder. There must be files within the folder. The files must start with the foldername. If you want slices, there must be `@` at the end of the basename. It is recommended(!) to have the `MachineOf=`, `Requires=` and `After=` options in the files. The file MUST be a systemd unitfile. 

All of this is the format we expect. If the user breaks that, stuff won't work

docs/terms.md
zeisss, line 26: > Why making a difference in wording only because the unit name changes?

Because semantics also change. Updating might behave differently. You cannot create a second or more slices.

xh3b4sd, line 26: Then name it and document as you think it is right. I am out in this topic.

zeisss, line 26: we have a name. Instance.

docs/README.md
xh3b4sd: Done

docs/README.md
xh3b4sd: Done

docs/terms.md
xh3b4sd: Done

docs/goal.md
xh3b4sd: Done

zeisss:
LGTM

xh3b4sd:
I am not sure the `instance` term discussion led to a sufficient conclusion. You stated something fishy in the documentation I wrote. Then we discussed and couldn't find a consensus. Then you decided there is something that is called `instance`, but what that means remains unclear. Unclear in my mental model. Unclear in the docs of this PR. Nevertheless you say LGTM. So what to do now?

zeisss:
> I am not sure the instance term discussion led to a sufficient conclusion

Agreed, but that shouldn't block us from getting this out.

> Then you decided there is something that is called instance

I stated what WE all have discussed before in the in the brainstorming sessions, how I understood it. I didn't want to give the impression I decided something.

Lets make an issue for the unclear definition of instance and discuss it together. Either in the issue or in person next week via a hangout or sthg.

xh3b4sd:
See https://github.com/giantswarm/inago/issues/60

PR #53, opened 2016-03-03
adds example

Nesurion:
adds example unit files


Nesurion:
@giantswarm/cena LG?

xh3b4sd:
Funny I thought about this in my PR https://github.com/giantswarm/formica/pull/52. Do we maybe want to document how it looks like when we are deploying this examples?

JosephSalisbury:
Yeah, a little example of how to get this going would be good

Nesurion:
Documenting it sounds good, but where would you do it? I think the options are :a: _inside the unit files_, :b: _in the readme/docs_

Also I think we should use one name troughout the files in the repo and the docs for the example. Currently the `terms.md` in #52 mentions a `mygroup`.
So two options aswell again:
:one: `example` / `example-main@.service`, `example-sidekick@.service`
:two: `mygroup` / `mygroup-foo@.service`, `mygroup-bar@.timer`

zeisss:
we need multiple examples anyway, so I would suggest an `examples` folder with a subfolder for each group,.

Nesurion:
@zeisss ack

Nesurion:
merged the example folder with the first `mygroup` example. @giantswarm/cena LG?

xh3b4sd:
Got the discussion about using the example lost? What was the decision?

Nesurion:
sorry forgot about that. I'd fear that when we put the output of inago commands using the examples into the unitfiles, no one will look there. So I'd say we document that with the other things inside the docs.
Vote: :b:

xh3b4sd:
Yeah, the unit files are no good place for this kind of documentation. 

I am thinking this way. We will have further examples. Then we probably don't want to add the same example workflow with the same commands executed for each example. What about simply merging this PR and adding one docs page where we show some example command output for each command. There the most basic example unit files can be used to execute these commands.

Nesurion:
sounds good

PR #54, opened 2016-03-04
adds status command output test

Nesurion:
adds tests to validate that the output printed by Ingao when running `ingaoctl status example`


Nesurion:
ping @giantswarm/cena

JosephSalisbury:
It might be an idea to try and squash these commits

.gitignore
```diff
@@ -1,3 +1,4 @@
 inagoctl
 .gobuild/
 coverage.txt
+profile.out
```
xh3b4sd, line 4: Where does this new file now come from?

Nesurion:
squashing published commits is kind of ugly. @giantswarm/cena are you ok with merging this this time an i'll reduce the number of commits in the future?

xh3b4sd:
> squashing published commits is kind of ugly.

Squashing commits is totally fine for me. It is only that we did not agree on doing this. The amount of commits within a PR is not a critical metric. It is more important that the history helps later in case you want to lookup information in the future. Squashing commits leads to a cleaner history but information loss of intermediate state. It is up to you.

Towards the tests in this PR. I miss one where the output needs to be expanded because the statuses differ within a group.

.gitignore
Nesurion, line 4: from: https://github.com/giantswarm/inago/blob/master/go.test.sh#L10
If the test panics the profile.out is not deleted, which might be fine, but I wouldn't want to commit it.

Nesurion:
ping @giantswarm/cena

zeisss:
Squashing commits in PR is fine for me too, if you work alone in the branch. Afaik we also said in our example-open-source repo's CONTRIBUTION.md file that this encouraged/expected?

cli/common_test.go
```diff
@@ -226,3 +229,160 @@ func Test_Common_createRequest(t *testing.T) {
...
+			Comment: "3 slices of 1 group with 2 units",
+			Expected: []string{
+				"Group | Units | FDState | FCState | SAState | IP | Machine",
+				"",
+				"example@1 | * | loaded | loaded | inactive | 172.17.8.101 | 505e0d7802d7439a924c269b76f34b5f",
+				"example@2 | * | loaded | loaded | inactive | 172.17.8.102 | 9ebb53b04b0d46fb94b4fd1b3f562d2b",
+				"example@3 | * | loaded | loaded | inactive | 172.17.8.103 | e3cb5f13a9164ba5b7eff6c920475e61",
+				"",
+			},
+			Input: input{
```
zeisss: can we put `Input` above `Expected` ? so its input, then output?

zeisss:
LGTM

PR #56, opened 2016-03-04
Rename formica to inago

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/547

```
$ git grep -i formica | wc -l
       0
$ git ls-files | grep -i formica | wc -l
       0
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
-formicactl
+inagoctl
 .gobuild/
 coverage.txt
--- a/README.md
+++ b/README.md
@@ -1,3 +1,3 @@
-# inago
+# Inago
 
 [![Build Status](https://api.travis-ci.org/giantswarm/inago.svg)](https://travis-ci.org/giantswarm/inago) [![codecov.io](https://codecov.io/github/giantswarm/inago/coverage.svg?branch=master)](https://codecov.io/github/giantswarm/inago?branch=master)
--- a/README.md
+++ b/README.md
@@ -1,3 +1,3 @@
-# formica
+# inago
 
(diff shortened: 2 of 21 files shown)


JosephSalisbury:
RFR @giantswarm/cena

README.md
```diff
@@ -1,3 +1,3 @@
-# formica
+# inago
```
xh3b4sd: Do we want to use title case? Inago is a name, so may `Inago` would be more appropriate when using it in free text.

JosephSalisbury: Yeah, that sounds good.

.gitignore
```diff
@@ -1,3 +1,3 @@
-formicactl
+inagoctl
```
zeisss, line 1: can we temporariyl leave the old one in? otherwise I guarantee someone will commit his old binary.

zeisss:
LGTM, although I would leave formicactl in the ignore list.

.gitignore
JosephSalisbury, line 1: I'd rather not, it means we'll have it hanging around. I'm sure we'll notice in the PR if someone is committing binaries.

common/common.go
```diff
@@ -1,5 +1,5 @@
 // Package common provides implementation of general interest for certain sub
-// packages of formica.
+// packages of inago.
```
xh3b4sd: Title case.

controller/status.go
```diff
@@ -171,7 +171,7 @@ var (
 )
 
 // AggregateStatus aggregates the given fleet and systemd states to a Status
-// known to formica based on the StatusIndex.
+// known to inago based on the StatusIndex.
```
xh3b4sd: Title case.

docs/README.md
```diff
@@ -1,5 +1,5 @@
 # docs
-Here we document important concepts and functionality formica supports.
+Here we document important concepts and functionality inago supports.
```
xh3b4sd: Title case.

.gitignore
xh3b4sd, line 1: +1 @JosephSalisbury

PR #59, opened 2016-03-04
Validation command

JosephSalisbury:
Fixes https://github.com/giantswarm/giantswarm/issues/523

### DIFF
--- a/validator/error.go
+++ b/validator/error.go
@@ -5,12 +5,12 @@ import (
 )
 
 var (
-	noUnitsInGroupError        = errgo.New("no units in group")
-	badUnitPrefixError         = errgo.New("unit does not have group prefix")
-	mixedSliceInstanceError    = errgo.New("group mixing scalable and non-scalable units")
-	multipleAtInGroupNameError = errgo.New("multiple @ symbols in group name")
-	multipleAtInUnitNameError  = errgo.New("multiple @ symbols in unit name")
-	unitsSameNameError         = errgo.New("unit named with same name as another unit")
+	noUnitsInGroupError       = errgo.New("no units in group")
+	badUnitPrefixError        = errgo.New("unit does not have group prefix")
+	mixedSliceInstanceError   = errgo.New("group mixing scalable and non-scalable units")
+	atInGroupNameError        = errgo.New("@ symbols in group name")
+	multipleAtInUnitNameError = errgo.New("multiple @ symbols in unit name")
+	unitsSameNameError        = errgo.New("unit named with same name as another unit")
 
 	groupsArePrefixError = errgo.New("group is prefix of another group")
 	groupsSameNameError  = errgo.New("group named with same name as another group")
@@ -31,9 +31,9 @@ func IsMixedSliceInstance(err error) bool {
 	return errgo.Cause(err) == mixedSliceInstanceError
 }
 
-// IsMultipleAtInGroupName returns true if the given error cause is multipleAtInGroupNameError.
-func IsMultipleAtInGroupName(err error) bool {
-	return errgo.Cause(err) == multipleAtInGroupNameError
+// IsAtInGroupNameError returns true if the given error cause is atInGroupNameError.
+func IsAtInGroupNameError(err error) bool {
+	return errgo.Cause(err) == atInGroupNameError
 }
 
 // IsMultipleAtInUnitName returns true if the given error cause is multipleAtInUnitNameError.
--- a/validator/error.go
+++ b/validator/error.go
@@ -5,9 +5,11 @@ import (
 )
 
 var (
-	noUnitsInGroupError     = errgo.New("no units in group")
-	badUnitPrefixError      = errgo.New("unit does not have group prefix")
-	mixedSliceInstanceError = errgo.New("group mixing scalable and non-scalable units")
+	noUnitsInGroupError        = errgo.New("no units in group")
+	badUnitPrefixError         = errgo.New("unit does not have group prefix")
+	mixedSliceInstanceError    = errgo.New("group mixing scalable and non-scalable units")
+	multipleAtInGroupNameError = errgo.New("multiple @ symbols in group name")
+	multipleAtInUnitNameError  = errgo.New("multiple @ symbols in unit name")
 
 	groupsArePrefixError = errgo.New("group is prefix of another group")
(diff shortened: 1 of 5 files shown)


validator/validator.go
```diff
@@ -0,0 +1,59 @@
...
+	for _, unit := range request.Units {
+		if !strings.HasPrefix(unit.Name, request.Group) {
+			return false, badUnitPrefixError
+		}
+	}
+
+	// Check that all units either have @ or they don't.
+	numUnitsWithAtSymbol := 0
+	for _, unit := range request.Units {
+		if strings.Contains(unit.Name, "@.") {
```
JosephSalisbury: I want a better way to do this, but I can't see it yet.

JosephSalisbury:
RFC @giantswarm/cena 

I'm still not certain on having this in a separate package or putting it into the controller package, or putting it into the cli. If there are any opinions, I'd like to hear them.

Also, `ok   _/home/travis/gopath/src/github.com/giantswarm/inago/validator  0.003s  coverage: 100.0% of statements` helloooooooo

JosephSalisbury:
Obviously, I haven't implemented the cli command yet, I want to get the core right first

validator/error.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"github.com/juju/errgo"
+)
+
+var (
+	noUnitsInGroupError     = errgo.New("no units in group")
+	badUnitPrefixError      = errgo.New("unit does not have group prefix")
+	mixedSliceInstanceError = errgo.New("group mixing scalable and non-scalable units")
+
+	groupsArePrefixError = errgo.New("group is prefix of another group")
```
xh3b4sd, line 15: We have the convention to create errors and their assertion methods. So the assertion methods are missing here.

validator/validator_test.go
```diff
@@ -0,0 +1,177 @@
...
+				controller.Request{
+					Group: "bat",
+				},
+				controller.Request{
+					Group: "batman",
+				},
+			},
+			valid: false,
+			err:   groupsArePrefixError,
+		},
```
xh3b4sd, line 207: I would like to see a test where both group names are indentical.

validator/validator_test.go
```diff
@@ -0,0 +1,177 @@
...
+				Group: "single",
+				Units: []controller.Unit{
+					controller.Unit{
+						Name: "bad-prefix.service",
+					},
+				},
+			},
+			valid: false,
+			err:   badUnitPrefixError,
+		},
```
xh3b4sd, line 158: I miss tests where group and unit names contain multiple `@` characters. IMO this should also not be valid.

xh3b4sd, line 158: Further there should be a test where unit names are completely equal. This may not be possible when relying on the file system, but this could happen later when there is network in between.

validator/validator_test.go
JosephSalisbury, line 207: Done, see https://github.com/giantswarm/inago/pull/59/files#diff-cf90a7e41f8ecad4a2c3c02e27b0dbf7R164

validator/validator_test.go
JosephSalisbury, line 158: Nice, all done

validator/validator.go
```diff
@@ -0,0 +1,82 @@
...
+			numUnitsWithAtSymbol++
+		}
+	}
+	if numUnitsWithAtSymbol > 0 && numUnitsWithAtSymbol < len(request.Units) {
+		return false, mixedSliceInstanceError
+	}
+
+	// Test there are not multiple @ symbols in the group name.
+	if strings.Count(request.Group, "@") > 1 {
```
zeisss: shouldn't the group name be free of `@` signs?

validator/validator.go
```diff
@@ -0,0 +1,82 @@
...
+
+	// Test there are not multiple @ symbols in any unit name.
+	for _, unit := range request.Units {
+		if strings.Count(unit.Name, "@") > 1 {
+			return false, multipleAtInUnitNameError
+		}
+	}
+
+	// Test that all unit names are unique
+	for i := 0; i < len(request.Units)-1; i++ {
```
zeisss: `for i, unit := range request.Units {`

no need for bording old-style for loops

validator/validator.go
```diff
@@ -0,0 +1,82 @@
...
+		if request.Units[i] == request.Units[i+1] {
```
zeisss: this test only verifies that there are no similar units next to each other, not in the whole list. You either need to sort the units first or put them into a map.

JosephSalisbury: yeah, i'm dumb <3

validator/validator.go
xh3b4sd: Things like this can be improved by separating the code into helper methods. Something like `unitsContain([]string, string) bool`. That way you can call it `if unitsContain(units, "@") { ... }`. In the helper method you just return true as soon as you see the string within a unit name. On the caller side you simply assert the condition against true or false. No counting necessary.

validator/validator_test.go
```diff
@@ -0,0 +1,231 @@
...
+			},
+			valid: false,
+			err:   multipleAtInUnitNameError,
+		},
+		{
+			request: controller.Request{
+				Group: "group",
+				Units: []controller.Unit{
+					controller.Unit{
+						Name: "group-unit1@.service",
```
xh3b4sd, line 143: This lines look pretty much the same. A comment would be nice to break the silence. Somebody may overlooks the difference of the unit names.

validator/validator.go
JosephSalisbury: Yeah, I did it this way so I can check pairwise without panicing - I don't get how to do that with your syntax

zeisss: oh yeah, the `i+1` access, mh. good question. I shut up now ^^

JosephSalisbury: <3

JosephSalisbury:
@giantswarm/cena RFR

I've half a mind to move the strings-related functions to somewhere better, but I don't see the point right now, if we're only using them here.

zeisss:
I think we want as much of the validation in the controller for all commands.

JosephSalisbury:
Okay, that sounds good to me - I'd like to add validation to other commands in another PR, though.

xh3b4sd:
LGTM

PR #61, opened 2016-03-04
use - instead of _ in doc file names

Nesurion:
@marians said to "prefer dash - over underscore _", so this branch replaces the _ with -


Nesurion:
ping @giantswarm/cena

zeisss:
LGTM

Issue #66, opened 2016-03-07
Add validation to other commands


JosephSalisbury:
It would be a good idea to integrate the validation into other commands, to make sure that invalid groups are not being submitted, for example.

PR #67, opened 2016-03-07
Add request validation to controller submit

JosephSalisbury:
Fixes https://github.com/giantswarm/inago/issues/66

### DIFF
--- a/cli/submit.go
+++ b/cli/submit.go
@@ -5,6 +5,8 @@ import (
 	"os"
 
 	"github.com/spf13/cobra"
+
+	"github.com/giantswarm/inago/validator"
 )
 
 var (
@@ -19,6 +21,12 @@ var (
 func submitRun(cmd *cobra.Command, args []string) {
 	req, err := createRequestWithContent(args)
 	if err != nil {
+		fmt.Printf("%#v\n", maskAny(err))
+		os.Exit(1)
+	}
+
+	ok, err := validator.ValidateRequest(req)
+	if !ok {
 		fmt.Printf("%#v\n", maskAny(err))
 		os.Exit(1)
 	}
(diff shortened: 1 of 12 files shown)


JosephSalisbury:
RFR @giantswarm/cena

cli/submit.go
```diff
@@ -23,6 +25,12 @@ func submitRun(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
+	ok, err := validator.ValidateRequest(req)
```
zeisss: don't we want this in the controller? So the controller validates it's input?

xh3b4sd: +1

controller/controller.go
```diff
@@ -113,65 +82,10 @@ type controller struct {
...
-	}
-
-	return newRequest, nil
-}
-
-func (c controller) Submit(req Request) (*task.Task, error) {
-	if len(req.Units) == 0 {
-		return nil, maskAnyf(invalidArgumentError, "units must not be empty")
+func (c controller) Submit(req api.Request) (*task.Task, error) {
+	ok, err := validator.ValidateRequest(req)
```
JosephSalisbury, line 86: This is the new validation

JosephSalisbury:
@giantswarm/cena RFR

I've had to pull out a bunch of stuff to get around the circular import _yay_

controller/controller.go
xh3b4sd, line 86: Why is the validator not simply part of the controller? Then we can prevent all this restructuring. Looks like validation is a controller capability. Just put the source of the validation package into the controller package. What do you think?

JosephSalisbury, line 86: Yeah, actually, that makes way more sense. I'll scrap this PR, go with that.

PR #69, opened 2016-03-08
Add validation to controller submit

JosephSalisbury:
Fixes https://github.com/giantswarm/inago/issues/66


JosephSalisbury:
@giantswarm/cena :fishing_pole_and_fish:

controller/controller_test.go
```diff
@@ -353,7 +353,7 @@ func TestController_Submit_Error(t *testing.T) {
 	// Assert
 	Expect(task).To(BeNil())
 	Expect(err).To(HaveOccurred())
-	Expect(err.Error()).To(Equal("invalid argument: units must not be empty"))
+	Expect(err.Error()).To(Equal("no units in group"))
```
Nesurion: is it possible to compare this to `noUnitsInGroupError` rather than the string? in case of changing the error message we would just change it in one place.

zeisss: `Expect(IsNoUnitsInGroup(err)).To(BeTrue())`

JosephSalisbury: Thanks gang <3

controller/error.go
```diff
@@ -73,3 +73,59 @@ var invalidArgumentError = errgo.Newf("invalid argument")
...
+func IsGroupsArePrefix(err error) bool {
+	return errgo.Cause(err) == groupsArePrefixError
+}
+
+var groupsSameNameError = errgo.New("group named with same name as another group")
+
+// IsGroupsSameName returns true if the given error cause is groupsSameNameError.
+func IsGroupsSameName(err error) bool {
+	return errgo.Cause(err) == groupsSameNameError
+}
```
xh3b4sd, line 131: The coverage here is 100%. See https://github.com/giantswarm/inago/blob/master/controller/error_test.go. If we care about coverage we should also test the new errors.

JosephSalisbury, line 131: I'm using the error assertions in the validator tests, so they are covered that way. I found it very useful to use the coverage to check all errors had tests against.

xh3b4sd, line 131: Right. I saw my browser extension was weird. The file is covered. All good.

xh3b4sd:
IIWFYLGTM

PR #71, opened 2016-03-08
fixed grouped status

xh3b4sd:
I noticed that grouping the status of groups does not work properly. This PR fixes the problem.

### DIFF
--- a/controller/status.go
+++ b/controller/status.go
@@ -79,7 +79,7 @@ func allStatesEqual(usl []fleet.UnitStatus) bool {
 					if m1.SystemdActive != m2.SystemdActive {
 						return false
 					}
-					if m1.UnitHash != m2.UnitHash {
+					if us1.Name == us2.Name && m1.UnitHash != m2.UnitHash {
 						return false
 					}
 				}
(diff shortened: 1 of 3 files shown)


controller/status.go
```diff
@@ -16,6 +16,8 @@ func (usl UnitStatusList) Group() ([]fleet.UnitStatus, error) {
 	matchers := map[string]struct{}{}
 	newList := []fleet.UnitStatus{}
 
+	hashesEqual := allHashesEqual(usl)
```
xh3b4sd, line 19: This is the trick. When we added unit hashes we compared the unit hashes of the unit files that were already grouped. Obviously it makes no sense to compare unit hashes with the units within the same group. We are interested in comparing the unit hashes across the slices.

xh3b4sd:
RFR @zeisss @Nesurion

controller/status.go
```diff
@@ -63,8 +67,27 @@ func groupUnitStatus(usl []fleet.UnitStatus, groupMember fleet.UnitStatus) ([]fl
...
 }
 
+// allHashesEqual is supposed to receive a list of unit statuses that is not
+// grouped. This is necessary to compare unit hashes across groups.
+func allHashesEqual(usl []fleet.UnitStatus) bool {
+	for _, us1 := range usl {
+		for _, us2 := range usl {
+			for _, m1 := range us1.Machine {
+				for _, m2 := range us2.Machine {
+					if common.UnitBase(us1.Name) == common.UnitBase(us2.Name) && m1.UnitHash != m2.UnitHash {
```
zeisss: can we move the `.Name` comparision two loops earlier?

zeisss:
LGTM, although I would change the check to do less looping.

controller/status.go
xh3b4sd: Done

PR #74, opened 2016-03-10
Logging

JosephSalisbury:
Fixes https://github.com/giantswarm/giantswarm/issues/536


JosephSalisbury:
@giantswarm/cena I would appreciate people's thoughts here.

My main idea has been to have one logger which we configure at some start time (here, in the cli package - in the future, during some server start up), and which other packages can grab as they need.
This is mainly so we can set log levels in one place, but also helps in terms of formatting.

I haven't committed any actual logging to keep my history clean, but it looks something like:

```
logger := logging.GetLogger()
logger.Debug(nil, "beep boop")
```

My main pain point is, because I need to wait for `cobra` to parse flags, I can't set up the logger earlier, meaning we can't use top level vars to provide the logger (e.g: having `var logger = logging.GetLogger()` at the top of a file, and forgetting about it)

If people can see a better way to initialise, that let's us do that, that would be great. The alternative is declaring a logger as needed.

xh3b4sd:
> My main pain point is, because I need to wait for cobra to parse flags, I can't set up the logger earlier

I don't understand the pain. We do this already. There is a global instance of e.g. the controller. This is configured in a cobra callback. This callback is fired after flags are parsed and before commands are executed. What do I miss?

logging/logging.go
```diff
@@ -0,0 +1,43 @@
...
+
+	// The default log level to use.
+	logLevel = "INFO"
+	// Whether to use color, by default.
+	useColor = false
+)
+
+// SetLogLevel sets the log level of the logger to be created.
+func SetLogLevel(level string) {
+	logLevel = level
```
xh3b4sd: I did not follow the logger discussions. What were the decisions? Why do we want this here? Why not simply glog? It has nice verbosity features. 

FYI I had this logging problem in one of my own projects as well recently and I wrote my own logger supporting filtering for log levels, verbosity and objects emitting the log messages.

JosephSalisbury:
With regards to: https://github.com/giantswarm/inago/pull/74#issuecomment-194819690

it's not really _pain_, it's more style: because the logger gets set up after cobra has parsed flags, any logger that is grabbed in `init()` (for example), won't respect the levels set with the flags.
This means we need to grab loggers "after" they've been set up, such as in functions themselves.

It's more that I can't decide if this is fine, or we want to be able to do something like:

```
var logger = logging.GetLogger()
```

at the top of a file once, and then use logger throughout.
We can't do that currently, we'd need to do something like:

```
func someFunction() {
   logger := logging.GetLogger()
   ...
   logger.Debug(nil, "i'm doing a thing")
}
```

If the latter is acceptable, or the way we should actually be doing this, then we're good :D

logging/logging.go
JosephSalisbury: @zeisss suggested using the `request-context` library. Our logging discussion on Slack seemed to lean towards using `go-logging` - which this library uses - but using the `request-context` library allows us to also use the context stuff it also has. This is mainly useful later when we move to client-server.

The `logging` package is just a little wrapper around it, mainly so we don't have to set up configs in multiple locations.

xh3b4sd:
> it's not really pain, it's more style: because the logger gets set up after cobra has parsed flags, any logger that is grabbed in `init()` (for example), won't respect the levels set with the flags.

Yes, this is a common problem we have. Every failure during initialization should simply `panic` or `fatal`.

logging/logging.go
xh3b4sd: Ok. Then we need to live with its limitations.

cli/inagoctl.go
```diff
@@ -59,6 +70,7 @@ func init() {
 	MainCmd.PersistentFlags().StringVar(&globalFlags.FleetEndpoint, "fleet-endpoint", "unix:///var/run/fleet.sock", "endpoint used to connect to fleet")
 	MainCmd.PersistentFlags().BoolVar(&globalFlags.NoBlock, "no-block", false, "block on synchronous actions or not")
 	MainCmd.PersistentFlags().BoolVarP(&globalFlags.Verbose, "verbose", "v", false, "verbose output or not")
+	MainCmd.PersistentFlags().BoolVar(&globalFlags.LogColor, "log-color", false, "whether to use colorful output for logs")
```
Nesurion: I dislike the extra flag here. Can't we detect if the output is a tty device and if so enable color, otherwise disable it.
https://github.com/mattn/go-isatty

JosephSalisbury:
@xh3b4sd okay, that seems positive. I'll stick with this style.

I'm planning on peppering a few debug and infos around to get us started on logging, so we'll see how it goes.

cli/inagoctl.go
JosephSalisbury: Done, i like this

Makefile
```diff
@@ -36,12 +36,15 @@ clean:
 	@builder get dep -b 7f60f83a2c81bc3c3c0d5297f61ddfa68da9d3b7 https://github.com/spf13/pflag.git $(GOPATH)/src/github.com/spf13/pflag
 	@builder get dep -b 983d3a5fab1bf04d1b412465d2d9f8430e2e917e https://github.com/ryanuber/columnize.git $(GOPATH)/src/github.com/ryanuber/columnize
 	@builder get dep -b e673fdd4dea8a7334adbbe7f57b7e4b00bdc5502 https://github.com/satori/go.uuid.git $(GOPATH)/src/github.com/satori/go.uuid
+	@builder get dep -b 51ed24df9dfd342d701d40b62c00ee5ad4826e45 https://github.com/giantswarm/request-context.git $(GOPATH)/src/github.com/giantswarm/request-context
```
zeisss: our own libs have tags. I would prefer them.

logging/logging.go
```diff
@@ -0,0 +1,43 @@
+// Package logging provides a logger, as well as configuration.
+package logging
```
zeisss, line 2: do we need another package? Can't this just be part of the CLI?

JosephSalisbury, line 2: I put it in a separate package so we were fine when we move to client-server.

And I prefer more, smaller packages.

JosephSalisbury:
Note: I chatted with Tim, and we agreed this global logger approach isn't great. I'm gonna pass the logger through to where it's required.

JosephSalisbury:
@giantswarm/cena RFR

major change is that i'm now passing the logger around properly, gotten rid of `logging` too - folded it into cli

cli/inagoctl.go
```diff
@@ -34,19 +38,34 @@ var (
...
+
+			logger := requestcontext.MustGetLogger(
+				requestcontext.LoggerConfig{
+					Name:                "inago",
+					Level:               logLevel,
+					Color:               isatty.IsTerminal(os.Stderr.Fd()),
+					IncludeNameInFormat: false,
+				},
+			)
+			newLogger = &logger
```
xh3b4sd: Why not directly assigning to `newLogger`?

JosephSalisbury: `cannot take the address of requestcontext.MustGetLogger(requestcontext.LoggerConfig literal)`

in go, you can't take the address of a function call, you need to assign it to a variable first (which you can then take the address of)

cli/inagoctl.go
```diff
@@ -34,19 +38,34 @@ var (
...
+
 			URL, err := url.Parse(globalFlags.FleetEndpoint)
 			if err != nil {
 				panic(err)
 			}
 
-			newFleetConfig := fleet.DefaultConfig()
+			newFleetConfig := fleet.DefaultConfig(newLogger)
```
xh3b4sd: -1 

The default config is there to provide some default configuration based on best effort of the packages knowledge. This should not receive any argument. Instead the config returned by `DefaultConfig` should be changed. See the line below where we modify the `Endpoint` property.

fleet/fleet_test.go
```diff
@@ -12,12 +13,21 @@ import (
 	"github.com/stretchr/testify/mock"
 )
 
+var logger = requestcontext.MustGetLogger(
```
xh3b4sd: I don't understand this. I have this in mind. Every object created provides a mechanism to create some default configuration that can be used to create a instance of itself. That means we have `DefaultFooConfig() FooConfig` and `NewFoo Foo`. Here `Foo` uses e.g. a logger. This logger is a property of `FooConfig`. `FooConfig` is initialized by best effort in `DefaultFooConfig`. So by default there should be some instance of a logger be set in the default config. That way we don't need to do this here, because the logger is already in the default config and we simply need to do `newFoo := foo.NewFoo(foo.DefaultFooConfig())` in most cases.

cli/inagoctl.go
xh3b4sd: Why do you even want to do this pointer redirection? Make `newLogger     *requestcontext.Logger` to `newLogger     requestcontext.Logger` and you are done.

fleet/fleet_test.go
zeisss: +1 the logger should be part of the Config

cli/inagoctl.go
```diff
@@ -34,6 +38,12 @@ var (
 			// This callback is executed after flags are parsed and before any
 			// command runs.
 
+			loggingConfig := logging.DefaultConfig()
+			if globalFlags.Verbose {
+				loggingConfig.LogLevel = "DEBUG"
+			}
```
xh3b4sd, line 45: Minor and not your fault. Here we see that required features do not match functionality of our internal implementations. Such Feature flips and jolty redirects often indicate some misconceptions.

cli/inagoctl.go
```diff
@@ -41,13 +51,20 @@ var (
...
 			}
 
+			newTaskServiceConfig := task.DefaultConfig()
+			newTaskServiceConfig.Logger = newLogger
+			newTaskService = task.NewTaskService(newTaskServiceConfig)
+
 			newControllerConfig := controller.DefaultConfig()
+			newControllerConfig.Logger = newLogger
 			newControllerConfig.Fleet = newFleet
+			newControllerConfig.TaskService = newTaskService
```
xh3b4sd, line 68: All this is so beautiful.

fleet/fleet.go
```diff
@@ -28,6 +29,9 @@ const (
 type Config struct {
 	Client   *http.Client
 	Endpoint url.URL
+
+	// Logger provides a logger.
```
xh3b4sd: The comments vary. #consistency

fleet/fleet.go
```diff
@@ -41,6 +45,7 @@ func DefaultConfig() Config {
 	newConfig := Config{
 		Client:   &http.Client{},
 		Endpoint: *URL,
+		Logger:   logging.NewLogger(logging.DefaultConfig()),
```
xh3b4sd, line 48: Beautiful.

logging/logging.go
```diff
@@ -0,0 +1,45 @@
...
+
+// Logger provides an interface for other loggers to implement.
+type Logger interface {
+	Debug(context Context, f string, v ...interface{})
+	Info(context Context, f string, v ...interface{})
+	Notice(context Context, f string, v ...interface{})
+	Warning(context Context, f string, v ...interface{})
+	Error(context Context, f string, v ...interface{})
+	Critical(context Context, f string, v ...interface{})
+}
```
xh3b4sd, line 19: Cool thing. Thanks.

logging/request_context_logger.go
```diff
@@ -0,0 +1,70 @@
+package logging
+
+import (
+	"github.com/giantswarm/request-context"
+)
+
+// NewRequestContextLogger returns a RequestContextLogger, given a Config.
+func NewRequestContextLogger(config Config) RequestContextLogger {
```
zeisss: why do we reimplement the whole request-context lib?

logging/request_context_logger.go
```diff
@@ -0,0 +1,70 @@
+package logging
```
xh3b4sd: Why two different files?

logging/request_context_logger.go
```diff
@@ -0,0 +1,70 @@
...
+	return rcl
+}
+
+// RequestContextLogger is a Logger based on the request-context library.
+type RequestContextLogger struct {
+	logger requestcontext.Logger
+}
+
+// contextToCtx takes a logging.Context, and returns a requestcontext.Ctx.
+func contextToCtx(context Context) requestcontext.Ctx {
```
xh3b4sd: For what do we need that?

logging/request_context_logger.go
JosephSalisbury: I didn't want to expose `request-context` outside of the interface. This wraps it all up.

cli/inagoctl.go
JosephSalisbury, line 45: I'm fine with this, for now.

logging/request_context_logger.go
JosephSalisbury: Arbitrary. Splits the interface and an implementation up. Made me feel good.

logging/request_context_logger.go
JosephSalisbury: requestcontext has `Ctx`, which I didn't want to expose. So, we have `logging.Context`, which we map to `Ctx`.

logging/request_context_logger.go
zeisss: then why do we use request-context at all? we could just call the logger here.

JosephSalisbury: Arguably, we could remove it.

I'm tempted not to, as there is some useful setup / formatting (see: https://github.com/giantswarm/request-context/blob/master/request_context.go#L46 and https://github.com/giantswarm/request-context/blob/master/request_context.go#L124), which I don't love the idea of basically copying.

JosephSalisbury:
I'm leaving this open for now, but starting next week I'm going to move these changes over to `request-context`, and work a little on sharpening that up.

After that, I'll revisit our logging here, and build it strictly on top of `request-context`.

cc @zeisss

JosephSalisbury:
Lol, I'm not doing that:
- we'll now use net.Context as our generic context implementation, instead of our own `logging.Context`.
- dropping dependency on `request-context`, and using `go-logging` directly.

JosephSalisbury:
@giantswarm/cena RFR

I'm gonna merge master after lunch, and sort out passing contexts into the logging in another PR

zeisss:
merge master please

JosephSalisbury:
ping @zeisss

task/task_service.go
```diff
@@ -103,7 +108,7 @@ func (ts *taskService) Create(action Action) (*Task, error) {
 		if err != nil {
 			_, markErr := ts.MarkAsFailedWithError(taskObject, err)
 			if markErr != nil {
-				fmt.Printf("[E] Task.MarkAsFailed failed: %#v\n", maskAny(markErr))
+				ts.Config.Logger.Error(nil, "[E] Task.MarkAsFailed failed: %#v", maskAny(markErr))
```
zeisss, line 111: do we want to use `context.TODO` for `nil` here? Or are we going to fix that anyway?

JosephSalisbury, line 111: I'm going to update everything to use `net.Context` in my next PR.

zeisss, line 111: LGTM then

PR #79, opened 2016-03-11
update integration tests

Nesurion:
Adds integration tests for the update command.


Nesurion:
@giantswarm/cena RFR

int-tests/002-update.t
```diff
@@ -0,0 +1,47 @@
+Create a group to play around with.
+  $ mkdir update-group
+  $ touch update-group/update-group-foo@.service
```
JosephSalisbury: Do you need the touches?

JosephSalisbury:
IIWFYLGTM

int-tests/002-update.t
zeisss: he shouldn't need them, afaict

PR #82, opened 2016-03-11
move status creation to controller

xh3b4sd:
For better test coverage and clearer responsibilities. I already hear Stephan. <3 display logic should not be in the controller. My thoughts are that this is only some mechanism providing a default structure how status can potentially be represented. This should not be in the CLI IMO. Real printing still happens in the CLI and this is good. We have several things in the controller that needs thinking in terms of restructuring. E.g. the request creation and the display structure. Anyway, this can be addressed separately. Nevertheless this PR should fake boost the coverage result. 

This is only an idea. We can talk about it or simply merge or close it. No strong feelings. Only wanted to do something.


xh3b4sd:
RFR @JosephSalisbury @zeisss @Nesurion

zeisss:
> My thoughts are that this is only some mechanism providing a default structure how status can potentially be represented.

Mh, but this is very much CLI dependent. If you want a website, you have to do it differently. If the goal of the tool becomes different, you have to do it differently. I see no benefit of moving this into the controller.  I would like to keep that package clean and reusable for everyone.

xh3b4sd:
> I see no benefit of moving this into the controller.

Merging #82 into master will increase coverage by +10.71%

> I would like to keep that package clean and reusable for everyone.

It still is.

zeisss:
> Merging #82 into master will increase coverage by +10.71%

Why does this increase coverage? Afaict you just moved code. Maybe we wrote the tests in  a way there were not picked up before?

> > I would like to keep that package clean and reusable for everyone.
> > It still is.

IMO no. I personally would actually move more stuff from the controller/ to the cli package, like the grouping logic, because that is also only relevant for the cli / the status command. 

What does @JosephSalisbury and @Nesurion think? :)

JosephSalisbury:
What about some kind of `Status` struct, which we can then do whatever we want with? `controller` produces a `Status`, `cli` is responsible for formatting it and printing it. 

You could argue that there is greater potential for re-use in the future, too.

xh3b4sd:
> Why does this increase coverage?

Coverage is calculated by the coverage that is created. Packages not being tested are not considered for calculation. We had one test in the cli package. This resulted in very low coverage, which in turn affected the overall coverage of the whole project negatively. I moved the test to some package that has a better coverage. That means I removed the negative coverage influence and turned it into a positive coverage influence. As the name of the feature branch state: `fake-test-coverage`.

> IMO no. I personally would actually move more stuff from the controller/ to the cli package, like the grouping logic, because that is also only relevant for the cli / the status command.

As already stated in the PR description. I think we agree that the state of the controller package is anyway not perfect. This can be improved in the next steps. This here is a intermediate step. Further, what means `IMO no`? What happened is this. Functionality to create a default structure is added. In the controller there already is something like this: `DefaultConig`. This also provides default structures. The one I added here makes no difference, but implements more logic to create the default structure. 

> What about some kind of Status struct, which we can then do whatever we want with? controller produces a Status, cli is responsible for formatting it and printing it.

I like the idea. Lets do this in the next steps.

zeisss:
> This can be improved in the next steps. 

Yes, but this step makes it worse (Yes, IMO only, as it seems, but thats why we have a discussion :) ) and I see no sense to move this now just to revert it on Monday. 

> Further, what means IMO no?

Not sure what you mean. IMO = I my opinion. I meant I disagree with your conclusion. 

> Functionality to create a default structure is added. In the controller there already is something like this: DefaultConig. This also provides default structures. The one I added here makes no difference, but implements more logic to create the default structure.

I do not see the relevance between default structures and and the generation of a UI output formation string.

xh3b4sd:
Fine. Closing.

PR #83, opened 2016-03-11
Show unloaded units too

zeisss:
When a unit is not loaded onto a host, e.g. when it has fleet metadata requirements that cannot be fulfilled, the groups/slices were not listed.

This PR fixes this, by rendering the machine fields with a dash in this case.

```
$ ./inagoctl --fleet-endpoint=http://172.17.8.101:49153 status alpha-app
Group          Units  FDState  FCState   SAState  IP  Machine

alpha-app@16d  *      loaded   inactive  -        -   -
alpha-app@f8d  *      loaded   inactive  -        -   -
```

### DIFF
--- a/cli/common.go
+++ b/cli/common.go
@@ -3,12 +3,14 @@ package cli
 import (
 	"bytes"
 	"fmt"
+	"net"
 	"os"
 	"regexp"
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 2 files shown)


zeisss:
ping @giantswarm/cena

JosephSalisbury:
LGTM

PR #85, opened 2016-03-11
Fix mygroup example to ensure its scheduled onto the same host

zeisss:
Ensures that all units of mygroup are scheduled on the same host.


zeisss:
ping @giantswarm/cena

JosephSalisbury:
LGTM

PR #86, opened 2016-03-11
Vagrant refactor

Nesurion:
removes the need to copy the `user-data.sample` to `user-data` by creating a copy before starting the test machine and removing it after the tests are done.


Nesurion:
@JosephSalisbury @zeisss what do you think?

zeisss:
LGTM

zeisss:
LGTM

PR #87, opened 2016-03-11
Move filesystem logic back to cli

zeisss:
This puts the logic to read the unit files for submit/update/validate into the client. In case the controller is separated from the CLI, there is no way for the controller to access the FS of the cli and would always need to upload the files to the controller. 

To not move us into a confusing situation that the same package reads files, gives them to a client which then passes them back (which someone could obviously see as a point of opzimization to just do direct), I would put this responsibility back to the cli package. It might make sense to move this into a subpackage though. Not sure.

### DIFF
--- a/cli/request.go
+++ b/cli/request.go
@@ -1,0 +1,47 @@
+package cli
+
+import (
+	"path/filepath"
+
+	"github.com/giantswarm/inago/controller"
+	"github.com/giantswarm/inago/file-system/spec"
+)
+
+// readUnitFiles reads the given dir and returns a map of filename => filecontent.
+// If any read operation fails, the error is immediately returned.
+func readUnitFiles(fs filesystemspec.FileSystem, dir string) (map[string]string, error) {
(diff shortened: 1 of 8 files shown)


zeisss:
ping @giantswarm/cena

cli/request.go
```diff
@@ -0,0 +1,47 @@
...
+func readUnitFiles(fs filesystemspec.FileSystem, dir string) (map[string]string, error) {
+	fileInfos, err := fs.ReadDir(dir)
+	if err != nil {
+		return nil, maskAny(err)
+	}
+
+	unitFiles := map[string]string{}
+	for _, fileInfo := range fileInfos {
+		if fileInfo.IsDir() {
+			continue
```
JosephSalisbury, line 21: This case isn't tested.

zeisss, line 21: It wasn't before, but agreed. I will add it.

JosephSalisbury:
Excluding small test case, LGTM

PR #93, opened 2016-03-14
Add contexts to logging, additional debug logging

JosephSalisbury:


JosephSalisbury:
@giantswarm/cena RFR

I haven't attached any values to the contexts, we can do that as and when we need to. I've also added a fair bit more debug logging.

zeisss:
LGTM

Issue #75, opened 2016-03-10
Command usage text is not consistent


JosephSalisbury:
e.g:

```
Available Commands:
  submit      submit a group
  status      status of a group
  start       Starts the specified group or slices
  stop        Stops the specified group or slices
  destroy     Destroys the specified group or slices
  validate    validate groups
```

The descriptions of the commands are not consistent.

I would recommend finding a standard (`etcd` might be a good start, or some other `cobra` based program), and then sticking to that.

PR #94, opened 2016-03-15
Make usage and flag help text a little more consistent

JosephSalisbury:
Fixes https://github.com/giantswarm/inago/issues/75
- Using upper case for both short and long command text
- Using lower case for flag text
- `<foo>` for required arguments
- `[foo]` for optional arguments
- `|` for or style
- `...` for "multiple" arguments

### DIFF
--- a/cli/destroy.go
+++ b/cli/destroy.go
@@ -10,9 +10,9 @@ import (
 
 var (
 	destroyCmd = &cobra.Command{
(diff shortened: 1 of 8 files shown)


JosephSalisbury:
@giantswarm/cena RFR

zeisss:
LGTM

PR #110, opened 2016-03-15
Removes wrong usage of atomic package

### LINKED ISSUE #109: We are using atomic.AddInt64 wrong
https://github.com/giantswarm/inago/blob/master/controller/update.go#L248

zeisss:
Fixes #109

### DIFF
--- a/controller/update.go
+++ b/controller/update.go
@@ -245,15 +245,13 @@ func (c controller) UpdateWithStrategy(ctx context.Context, req Request, opts Up
 			}
 			if ok {
 				go func() {
-					v := atomic.AddInt64(&addInProgress, 1)
-					addInProgress = v
+					atomic.AddInt64(&addInProgress, 1)
 					err := c.addFirst(ctx, newReq, opts)
 					if err != nil {
 						fail <- maskAny(err)
 						return
 					}
-					v = atomic.AddInt64(&addInProgress, -1)
-					addInProgress = v
+					atomic.AddInt64(&addInProgress, -1)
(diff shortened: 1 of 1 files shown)


zeisss:
ping @giantswarm/cena

JosephSalisbury:
LGTM

controller/update.go
```diff
@@ -268,15 +266,13 @@ func (c controller) UpdateWithStrategy(ctx context.Context, req Request, opts Up
...
-					removeInProgress = v
+					atomic.AddInt64(&removeInProgress, 1)
 					err := c.removeFirst(ctx, newReq, opts)
 					if err != nil {
 						fail <- maskAny(err)
 						return
 					}
-					v = atomic.AddInt64(&removeInProgress, -1)
-					removeInProgress = v
+					atomic.AddInt64(&removeInProgress, -1)
```
xh3b4sd, line 275: What was actually wrong here? I did it this weird way because go vet or something complained about this.

zeisss, line 275: by assigning the variable again you have concurrent writes again and the whole usage of `atomic` is useless.

xh3b4sd, line 275: Makes sense. Anyway I am really confused about this case and I looked that up what to do about it.

zeisss, line 275: Not sure what you mean - do you want more explanation?

xh3b4sd, line 275: I wan to remember the case I had to comprehend what happened and why it now is like I originally wanted to have it. Nothing to do for you here. Thanks.

Issue #106, opened 2016-03-15
Automate integration tests


zeisss:
TB: 1d

Req
- have a separate status in Github PRs
- Have the output of tests available (link?)

Ideas
- Run on a separate ec2 machine, put private ssh key into travis or whatever?

zeisss:
Followup #145 #146

PR #111, opened 2016-03-15
Automated integration tests

JosephSalisbury:
Fixes https://github.com/giantswarm/inago/issues/106

Todo:
- [x] Test out Travis build matrix
- [x] Set up EC2 box, test script
- [x] Add new build script to launch either unit or integration tests
- [x] Rebuild SSH key and server
- [x] Document server set up
- [x] Clean up fabric output
- [x] Fix integration tests
- [x] Make integration tests not interfere with each other on the same cluster / cleanup after themselves - remove global fleet unit destruction
- [x] Look into encrypting SSH key
- [x] Add regenerated SSH key to keepass
- [x] Put in ticket for launching new machines on each integration test


JosephSalisbury:
I've removed the update integration test as I was having serious issues getting it to work. Secondly, although the int test is green, there's definite flapping.

JosephSalisbury:
@giantswarm/cena RFR

JosephSalisbury:
Note: I have limited travis to 1 concurrent job

.travis.yml
```diff
@@ -6,29 +6,28 @@ go:
...
+        - ./builder-cache
 
 notifications:
     email: false
 
 before_install:
-- cp ./builder-cache/builder-$BUILDER_VERSION ./builder || wget https://downloads.giantswarm.io/builder/$BUILDER_VERSION/builder && chmod +x ./builder
-- export PATH=$PATH:$PWD
-- go get -u github.com/golang/lint/golint
+    - openssl aes-256-cbc -K $encrypted_314659c9f31c_key -iv $encrypted_314659c9f31c_iv -in inago-integration-test.pem.enc -out ./inago-integration-test.pem -d
```
zeisss: can you add some comments what you do here and maybe add a simple existence test before?

```
test -z "encrypted_314659c9f31c_key" && echo "No \$encrypted_314659c9f31c_key given. Mandatory." && exit 1
```

or sthg

.travis.yml
```diff
@@ -6,29 +6,28 @@ go:
...
+    - chmod +x ./builder
+    - export PATH=$PATH:$PWD
+    - go get -u github.com/golang/lint/golint
 
 script:
-- make lint
-- make ci-test
-- make ci-build
-- ./inagoctl
+    - ./ci.sh
```
zeisss, line 33: didn't you opt against this in the past? Whats the reason for doing this with the `matrix` option above now?

ci.sh
```diff
@@ -0,0 +1,19 @@
...
+    bash <(curl -s https://codecov.io/bash)
+elif [ $TEST_SUITE == "integration" ]; then
+    pip install --user fabric
+    make ci-build
+    eval "$(ssh-agent -s)"
+    chmod 400 ./inago-integration-test.pem
+    ssh-add ./inago-integration-test.pem 2>/dev/null
+    fab run_int_test
+else
+    echo "Unknown test suite"
```
zeisss, line 17: print out the `$TEST_SUITE` here please

docs/integration-server-setup.md
```diff
@@ -0,0 +1,12 @@
+# Integration Test Server Setup
+
+This document describes the set up for the integration test server.
```
zeisss, line 3: we should automate setting that up at some point too :dancer:

JosephSalisbury, line 3: https://github.com/giantswarm/inago/issues/145

ci.sh
JosephSalisbury, line 17: This is already done by Travis, e.g: https://travis-ci.org/giantswarm/inago/jobs/116946691#L137

.travis.yml
JosephSalisbury, line 33: I don't love it, but we need a conditional over `TEST_SUITE`. This means that we launch two jobs, due to the matrix, and the script then decides what commands to run based on `TEST_SUITE`. Can't do that in just the script section.

int-tests/002-update.t
```diff
@@ -1,45 +0,0 @@
-Create a group to play around with.
```
zeisss, line 1: why was this deleted?

.travis.yml
JosephSalisbury: I'll add a comment. `openssl` will break if the key isn't present, so I don't see the point of the test.

.travis.yml
zeisss, line 33: And this is just so it's faster?

int-tests/002-update.t
JosephSalisbury, line 1: Because it was super borked on my branch, and it's still present in your branch. My plan was to merge this, and then merge yours.

JosephSalisbury, line 1: Ah, but you just merged yours already.

.travis.yml
zeisss: Lets quote the variables then, so they get passed as an empty string argument if not present at least.

.travis.yml
JosephSalisbury, line 33: yes, this means the unit and integration tests run in parallel

int-tests/002-update.t
zeisss, line 1: can you bring that file back?

zeisss:
Bring back the deleted test, then LGTM

PR #113, opened 2016-03-15
Return taskObject.Error for failed tasks

zeisss:
We are doing it right [here](https://github.com/giantswarm/inago/blob/master/controller/update.go#L111) but wrong everywhere else. this makes it consistent.

### DIFF
--- a/controller/update.go
+++ b/controller/update.go
@@ -10,13 +10,13 @@ import (
 	"github.com/giantswarm/inago/task"
 )
 
-func (c controller) executeTaskAction(f func(req Request) (*task.Task, error), req Request) error {
-	taskObject, err := f(req)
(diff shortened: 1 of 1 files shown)


zeisss:
ping @giantswarm/cena

controller/update.go
```diff
@@ -10,6 +10,22 @@ import (
...
+	if err != nil {
+		return maskAny(err)
+	}
+	closer := make(<-chan struct{})
+	taskObject, err = c.WaitForTask(taskObject.ID, closer)
+	if err != nil {
+		return maskAny(err)
+	}
+	if task.HasFailedStatus(taskObject) {
+		return maskAny(fmt.Errorf(taskObject.Error))
```
zeisss, line 24: this is the actual fix, which we didn't do everywhere. See first commit. I then noticed the code duplication and cleaned that up too. Thus the title is a little bit misleading.

JosephSalisbury:
LGTM

Issue #91, opened 2016-03-14
start blocks forever for global units with no slices


zeisss:
With #90 we can now submit global units (see examples/). But starting them, blocks the cli forever (I guess).

zeisss:
Status works though.

```
kallisto:example zeisss$ ../inagoctl --fleet-endpoint="http://172.17.8.101:49153" status simple-global-unit
Group               Units  FDState   FCState   SAState  IP            Machine

simple-global-unit  *      launched  inactive  active   172.17.8.101  4ea5734c9d6347829d9236502b79c6ca
simple-global-unit  *      launched  inactive  active   172.17.8.102  52f3c1e0193243d8aedbf54052f83c84
simple-global-unit  *      launched  inactive  active   172.17.8.103  99f23dbedf2f4bdd9ed9f2600b5944c3

kallisto:example zeisss$ ../inagoctl --fleet-endpoint="http://172.17.8.101:49153" status -v simple-global-unit
Group               Units                                  FDState   FCState   SAState  Hash                                      IP            Machine

simple-global-unit  simple-global-unit-replicator.service  launched  inactive  active   c26304a69bc42fd4ea7f804618ce628fa889a1de  172.17.8.101  4ea5734c9d6347829d9236502b79c6ca
simple-global-unit  simple-global-unit-replicator.service  launched  inactive  active   c26304a69bc42fd4ea7f804618ce628fa889a1de  172.17.8.102  52f3c1e0193243d8aedbf54052f83c84
simple-global-unit  simple-global-unit-replicator.service  launched  inactive  active   c26304a69bc42fd4ea7f804618ce628fa889a1de  172.17.8.103  99f23dbedf2f4bdd9ed9f2600b5944c3

```

PR #116, opened 2016-03-16
Fix global units hanging on start

JosephSalisbury:
Fixes https://github.com/giantswarm/inago/issues/91

### DIFF
--- a/controller/status.go
+++ b/controller/status.go
@@ -213,6 +213,13 @@ var (
 	StatusIndex = []StatusContext{
 		{
 			FleetCurrent:  "inactive",
+			FleetDesired:  "launched",
+			SystemdActive: "active",
+			SystemdSub:    "running",
+			Aggregated:    StatusRunning,
+		},
+		{
+			FleetCurrent:  "inactive",
 			FleetDesired:  "*",
 			SystemdActive: "*",
 			SystemdSub:    "*",
--- a/controller/status_test.go
+++ b/controller/status_test.go
@@ -295,6 +295,14 @@ func Test_Status_AggregateStatus(t *testing.T) {
 			ErrorMatcher: IsInvalidUnitStatus,
 			Expected:     "",
 		},
+		{
+			FC:           "inactive",
+			FD:           "launched",
+			SA:           "active",
+			SS:           "running",
+			ErrorMatcher: nil,
+			Expected:     StatusRunning,
+		},
 	}
 
 	for i, testCase := range testCases {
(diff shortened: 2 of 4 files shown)


task/task_service.go
```diff
@@ -196,16 +200,24 @@ func (ts *taskService) WaitForFinalStatus(ctx context.Context, taskID string, cl
...
+				ts.Config.Logger.Debug(ctx, "task: has final status: %#v", taskObject)
 				return taskObject, nil
 			}
+
+			ts.Config.Logger.Debug(ctx, "task: does not have final status: %#v", taskObject)
 		}
 	}
+
+	ts.Config.Logger.Error(ctx, "task: waiting has fallen through")
+	return nil, nil
```
zeisss: if the return wasn't required before, why should this code now execute?

JosephSalisbury: Because I was getting paranoid :smile:

controller/status.go
```diff
@@ -213,6 +213,13 @@ var (
 	StatusIndex = []StatusContext{
 		{
 			FleetCurrent:  "inactive",
+			FleetDesired:  "launched",
+			SystemdActive: "active",
+			SystemdSub:    "running",
+			Aggregated:    StatusRunning,
+		},
```
JosephSalisbury, line 220: This is the actual fix - before, the `StatusContext` beneath was matched for this case, which lead `inago` to believe the aggregated status was `StatusStopped`, which was incorrect. This lead to repeatedly waiting for the status to become running.

JosephSalisbury:
@giantswarm/cena RFR

JosephSalisbury:
I would also like a WFM

controller/status.go
zeisss, line 220: Shouldn't we rather change this to `*`? https://github.com/giantswarm/inago/blob/fix-global-start-hang/controller/status.go#L271

JosephSalisbury, line 220: Sorry, I don't follow. "this" == ?

zeisss, line 220: this = the line in the URL. This is just broken because that line doesn't contain inactive. As we more or less want to ignore the fleet state anyway, we could put a `*` there, instead of adding a new case here.

zeisss:
WFM

controller/status_test.go
```diff
@@ -295,6 +295,14 @@ func Test_Status_AggregateStatus(t *testing.T) {
...
 			Expected:     "",
 		},
+		{
+			FC:           "inactive",
+			FD:           "launched",
+			SA:           "active",
+			SS:           "running",
+			ErrorMatcher: nil,
+			Expected:     StatusRunning,
+		},
```
xh3b4sd, line 305: Is this actually a test for the former broken behaviour? If so I would like to add a comment pointing to this PR and its discussion so one knows about the importance of this special case.

JosephSalisbury, line 305: Don't worry, I just did it <3

PR #117, opened 2016-03-16
Validate improvement

Nesurion:
- [x] don't validate folder without unit files
- [x] help text shows that multiple group names can be provided
- [x] use print instead of logger
- [x] show all errors / warnings, instead of exiting after the first occurrence

### DIFF
--- a/cli/validate.go
+++ b/cli/validate.go
@@ -38,7 +38,7 @@ func validateRun(cmd *cobra.Command, args []string) {
 				// If the directory is empty, don't validate it
 				subfiles, err := ioutil.ReadDir(file.Name())
 				if err != nil {
-					newLogger.Critical(newCtx, "%#v", maskAny(err))
+					newLogger.Error(newCtx, "%#v", maskAny(err))
 					os.Exit(1)
 				}
 				if len(subfiles) == 0 {
--- a/cli/validate.go
+++ b/cli/validate.go
@@ -72,7 +72,7 @@ func validateRun(cmd *cobra.Command, args []string) {
 		if ok {
 			fmt.Printf("Group '%v' is valid.\n", request.Group)
 		} else {
-			fmt.Printf("Group '%v' not valid: %v.\n", request.Group, err)
+			fmt.Printf("Group '%v' not valid: %v", request.Group, err)
 		}
 	}
 
--- a/cli/validate.go
+++ b/cli/validate.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"sort"
@@ -58,16 +59,16 @@ func validateRun(cmd *cobra.Command, args []string) {
 	for _, request := range requests {
 		ok, err := controller.ValidateRequest(request)
 		if ok {
-			newLogger.Info(newCtx, "Group '%v' is valid.", request.Group)
+			fmt.Printf("Group '%v' is valid.\n", request.Group)
 		} else {
-			newLogger.Info(newCtx, "Group '%v' not valid: %v.", request.Group, err)
+			fmt.Printf("Group '%v' not valid: %v.\n", request.Group, err)
 		}
 	}
 
 	ok, err := controller.ValidateMultipleRequest(requests)
 	if ok {
-		newLogger.Info(newCtx, "Groups are valid globally.")
+		fmt.Println("Groups are valid globally.")
 	} else {
-		newLogger.Info(newCtx, "Groups are not valid globally:", err)
+		fmt.Printf("Groups are not valid globally: %v\n", err)
 	}
 }
--- a/cli/validate.go
+++ b/cli/validate.go
@@ -35,6 +35,17 @@ func validateRun(cmd *cobra.Command, args []string) {
 
 		for _, file := range files {
 			if file.IsDir() && !strings.HasPrefix(file.Name(), ".") {
+				// If the directory is empty, don't validate it
+				subfiles, err := ioutil.ReadDir(file.Name())
+				if err != nil {
+					newLogger.Critical(newCtx, "%#v", maskAny(err))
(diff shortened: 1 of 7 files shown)


cli/validate.go
```diff
@@ -58,16 +70,16 @@ func validateRun(cmd *cobra.Command, args []string) {
 	for _, request := range requests {
 		ok, err := controller.ValidateRequest(request)
 		if ok {
-			newLogger.Info(newCtx, "Group '%v' is valid.", request.Group)
+			fmt.Printf("Group '%v' is valid.\n", request.Group)
```
zeisss, line 73: this reverses what Joe just added. I personally like it though - but is it intended? :dancer:

Nesurion, line 73: yes its intended. @joe and I talked and we feel that printing here is the nicer output.

controller/validator.go
```diff
@@ -85,14 +85,15 @@ func ValidateSubmitRequest(request Request) (bool, error) {
 // ValidateRequest takes a Request, and returns whether it is valid or not.
 // If the request is not valid, the error provides more details.
 func ValidateRequest(request Request) (bool, error) {
+	var error ValidationError
```
zeisss: lets not call this `error`

cli/validate.go
```diff
@@ -13,7 +14,7 @@ import (
 
 var (
 	validateCmd = &cobra.Command{
-		Use:   "validate [directory]",
+		Use:   "validate [directory...]",
```
JosephSalisbury, line 17: <3

cli/validate.go
```diff
@@ -34,6 +35,17 @@ func validateRun(cmd *cobra.Command, args []string) {
 
 		for _, file := range files {
 			if file.IsDir() && !strings.HasPrefix(file.Name(), ".") {
+				// If the directory is empty, don't validate it
+				subfiles, err := ioutil.ReadDir(file.Name())
+				if err != nil {
+					newLogger.Critical(newCtx, "%#v", maskAny(err))
```
JosephSalisbury: I'd use `Error` here, instead of `Critical`. I think `Critical` should be reserved for "the nukes are launching" level stuff.

Nesurion: Well after hitting the problem and reporting it, we exit 1. 
For me error is like: shit went down, but maybe we'll manage to fix it.
Critical: gone so south that we're all gonna die in a second.

controller/error.go
```diff
@@ -6,6 +6,22 @@ import (
...
+// ValidationError capsules validation errors into one error struct.
+// It is returned when the validation fails. causingErrors contains all
+// errors that occurenced, while validating the request.
+type ValidationError struct {
+	causingErrors []error
+}
+
+func (e ValidationError) Error() string {
+	// TODO better error message
+	msg := "group or unit invalid:\n"
```
Nesurion: I'd like to find a better formatting here. An output might look like this currently:

```
inagoctl validate
Group 'invalid' not valid: group or unit invalid:
    group mixing scalable and non-scalable units
    unit does not have group prefix
```

I like to tabs, but redundant "group or unit invalid" is not nice. Anyone got an idea how to get a nicer message here?

zeisss: How about we return the group as part of the validationError and print the errors by group?

```
Validation Errors found:
 * group 'mygroup': mixes scalable and unscalable units
 * group 'mygroup': wrong prefix
 * group 'global-simple-unit': wrong prefix
```

zeisss: also the formatting is a CLI problem, please move the code there.

Nesurion:
@giantswarm/cena RFR

cli/validate.go
JosephSalisbury: We're using `Error` in all other cases in the `cli` package for this pattern. It would be more consistent.

controller/error.go
```diff
@@ -6,6 +6,36 @@ import (
 	"github.com/juju/errgo"
 )
 
+// ValidationError capsules validation errors into one error struct.
+// It is returned when the validation fails. causingErrors contains all
```
JosephSalisbury: `causingErrors` should be `CausingErrors`

controller/controller.go
```diff
@@ -192,11 +192,9 @@ func (c controller) GroupNeedsUpdate(req Request) (Request, bool, error) {
 
 func (c controller) Submit(ctx context.Context, req Request) (*task.Task, error) {
 	c.Config.Logger.Debug(ctx, "controller: handling submit")
-
 	if ok, err := ValidateSubmitRequest(req); !ok {
-		return nil, maskAny(err)
+		return nil, err
```
zeisss: Why was this removed?

controller/error.go
```diff
@@ -6,6 +6,36 @@ import (
...
+	return msg
+}
+
+// Add adds the given error to the list of validation errors
+func (e *ValidationError) Add(err error) {
+	e.CausingErrors = append(e.CausingErrors, err)
+}
+
+// Contains returns true if the given error is present in the ValidationError
+func (e *ValidationError) Contains(err error) bool {
```
zeisss: can we pass a checker in here? So `Contains( f func(error) bool) bool` ?

controller/controller.go
Nesurion: masking the error results in a failed test: 

```
panic: interface conversion: error is *errgo.Err, not *controller.ValidationError [recovered]
    panic: interface conversion: error is *errgo.Err, not *controller.ValidationError
```

maskAny changes the error type to *errgo.Err. In order the cast the err to *controller.ValidationError, the submit func may not change the type.

zeisss: Use `errgo.Cause()` then please. The `maskAny` adds a stack trace and we do this everywhere.

Nesurion:
ping @giantswarm/cena

zeisss:
LGTM

Issue #115, opened 2016-03-16
Print task IDs from context nicer (and all the other things in the context)


JosephSalisbury:
While debugging, I found I was printing the task IDs in a lot of logging statements. It would be good to add the task ID to the context as early as possible, so it gets passed through automatically.

zeisss:
With #118 we add the ID to the context, but it is still not printed.

zeisss:
actually it is also printed, but it looks ugly and does not overwrite each other but gets appended somehow.

xh3b4sd:
I call for comments. Who has strong feelings about this? How should it look like? Please provide more precise information on what "Print task IDs from context nicer" actually means.

JosephSalisbury:
I'd like it if we could dump the "WithValue" crap, and shorten it when it's just the Background context.

xh3b4sd:
I have no idea how to do so. There is no way to fetch all information a context holds.

xh3b4sd:
Can somebody came up with a sketch? How should log line look like? Lets get design ideas right.

zeisss:
we could add a way to register the keys we are interested in and print only those.

JosephSalisbury:
taking 

```
2016-04-18 10:12:01.765 | DEBUG    | context.Background: controller: handling submit
2016-04-18 10:12:01.808 | DEBUG    | context.Background.WithValue("task-id", "dc640342-95cd-4819-b3c8-8ae298b5ab0a"): controller: fetching group status
```

as two random log lines to play with, we could have:

```
2016-04-18 10:12:01.765 | DEBUG    | controller: handling submit
2016-04-18 10:12:01.808 | DEBUG    | task-id:dc640342-95cd-4819-b3c8-8ae298b5ab0a | controller: fetching group status
```

basically, removing the `context.Background` foo, making the value part smaller.

PR #118, opened 2016-03-16
Put the task-id in the context when executing tasks

zeisss:

### DIFF
--- a/task/task_service.go
+++ b/task/task_service.go
@@ -9,6 +9,11 @@ import (
 	"github.com/giantswarm/inago/logging"
 )
 
+const (
+	// ContextTaskID is the key for the current task-id stored in the context.Context when executing tasks.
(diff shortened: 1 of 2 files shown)


zeisss:
ping @giantswarm/cena

JosephSalisbury:
Does this fix https://github.com/giantswarm/inago/issues/115?

JosephSalisbury:
But yes, LGTM

zeisss:
uh, I didn't see #115 :D it is part of it, yes. But thsi still doesn't trigger the ID getting logged.

JosephSalisbury:
I think the issue is that a new context is returned, so the "same" context, but in a higher scope isn't affected.

zeisss:
I know, that's why I overwrite `ctx` in the code. so all following calls inside the scope are affected, but not the client.

PR #120, opened 2016-03-16
fleet: For global units, return a nicer global status

zeisss:
This adds special treatment for global units.  Fleet returns `"inactive"` for the CurrentStatus, even if all units on all machines are running just fine. This PR changes the returned FleetCurrentStatus to always be FleetDesiredStatus, to make this a bit more logical.

### DIFF
--- a/fleet/fleet.go
+++ b/fleet/fleet.go
@@ -4,6 +4,7 @@
 package fleet
 
 import (
+	"strings"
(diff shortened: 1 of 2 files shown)


JosephSalisbury:
LGTM

zeisss:
ping @JosephSalisbury I added tests.

PR #123, opened 2016-03-16
IntTests: Golden path for instance groups

zeisss:
This PR adds an int test to run the submit/Start/stop/destroy golden path for a simple instance group (no slices) with two units. It seem to basically work, although if I do not redirect the output, it does not yet look like what I expect. I guess we have a bug in inago here. But I will focus on that in a separate PR.


zeisss:
ping @giantswarm/cena

JosephSalisbury:
does this actually test anything?

zeisss:
Yes, the commands need to return success.

JosephSalisbury:
cool, LGTM

PR #126, opened 2016-03-16
Add validation for update flags

JosephSalisbury:
Fixes https://github.com/giantswarm/inago/issues/105

### DIFF
--- a/controller/update.go
+++ b/controller/update.go
@@ -201,6 +201,16 @@ func (c controller) UpdateWithStrategy(ctx context.Context, req Request, opts Up
 	var addInProgress int64
 	var removeInProgress int64
 
+	// To keep n units alive during the update,
+	// we need to have just more than that in current units,
+	// and units we're allowed to create during the update.
+	if !(numTotal+opts.MaxGrowth > opts.MinAlive) {
+		return maskAnyf(
+			updateNotAllowedError,
+			"not enough current or growth units to satisfy liveness requirements",
+		)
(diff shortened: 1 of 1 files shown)


JosephSalisbury:
@giantswarm/cena RFR

I've tested this manually, and the validation seems to work. I've opened https://github.com/giantswarm/inago/issues/131 to address the lack of unit tests for update.

controller/update.go
```diff
@@ -201,6 +201,16 @@ func (c controller) UpdateWithStrategy(ctx context.Context, req Request, opts Up
 	var addInProgress int64
 	var removeInProgress int64
 
+	// To keep n units alive during the update,
+	// we need to have just more than that in current units,
+	// and units we're allowed to create during the update.
+	if !(numTotal+opts.MaxGrowth > opts.MinAlive) {
```
zeisss, line 207: this allows me to do this:

```
inagoctl submit mygroup 5
instalctl start mygroup
inagoctl update mygroup --max-growth=5 --min-alive=9
```

Which clearly means, it cannot reduce the number of slices after the initial growth. I don't think this is what we want.

JosephSalisbury, line 207: Yeah, this doesn't make sense. I'll have a think.

JosephSalisbury:
I'm closing this for now - https://github.com/giantswarm/inago/pull/126#discussion_r56483602 shows a fairly fundamental issue.

PR #133, opened 2016-03-17
Log a warning if multiple status contexts matched during aggregation

JosephSalisbury:
Fixes https://github.com/giantswarm/inago/issues/119

### DIFF
--- a/controller/status.go
+++ b/controller/status.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/giantswarm/inago/common"
 	"github.com/giantswarm/inago/fleet"
+	"github.com/giantswarm/inago/logging"
 )
 
 // UnitStatusList represents a list of UnitStatus.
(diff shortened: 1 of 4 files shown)


JosephSalisbury:
@giantswarm/cena RFR

controller/status.go
```diff
@@ -277,6 +263,11 @@ var (
 	}
 )
 
+// Aggregator provides methods for aggregating status.
+type Aggregator struct {
```
zeisss, line 267: either make the functions on this type public as well or hide the type. A public type with no functions is useless.

zeisss:
LGTM otherwise

controller/status.go
JosephSalisbury, line 267: https://github.com/giantswarm/inago/pull/133/files#diff-9155f40475e332d8aee80a437105af6bR279 AggregateStatus is public

controller/status.go
```diff
@@ -285,29 +276,55 @@ var (
...
+	}
+
+	numAggregatedStatuses := len(aggregatedStatuses)
+
+	if numAggregatedStatuses == 0 {
+		return "", maskAnyf(invalidUnitStatusError, "fc: %s, fd: %s, sa: %s, ss: %s", fc, fd, sa, ss)
+	}
+
+	if numAggregatedStatuses > 1 {
+		a.Logger.Warning(nil, "Multiple statuses found for: fc: %s, fd: %s, sa: %s, ss: %s", fc, fd, sa, ss)
```
xh3b4sd, line 306: Do we know when this happens? Neither the issue nor the PR provide some helpful information.

JosephSalisbury, line 306: This happened and was fixed in https://github.com/giantswarm/inago/pull/116 fixed it - the warning has been kept in the case that it happens again.

PR #148, opened 2016-03-21
Authenticate with googlesource to avoid ratelimiting

JosephSalisbury:

### DIFF
--- a/gitcookies.sh
+++ b/gitcookies.sh
@@ -1,0 +1,9 @@
+touch ~/.gitcookies
+chmod 0600 ~/.gitcookies
+
+git config --global http.cookiefile ~/.gitcookies
+
+tr , \\t <<\__END__ >>~/.gitcookies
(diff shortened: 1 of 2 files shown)


gitcookies.sh
```diff
@@ -0,0 +1,9 @@
+touch ~/.gitcookies
+chmod 0600 ~/.gitcookies
+
+git config --global http.cookiefile ~/.gitcookies
+
+tr , \\t <<\__END__ >>~/.gitcookies
+go.googlesource.com,FALSE,/,TRUE,2147483647,o,git-joseph.giantswarm.io=1/vxlFq4LObBrj22Iiy5AKT3g206S0oYp-lMhS73s673Q
```
zeisss, line 7: I am not sure we want this plaintext in github.

JosephSalisbury, line 7: We don't, I 100% meant to encrypt this.

JosephSalisbury:
I have revoked these credentials

Issue #147, opened 2016-03-21
Fix go get issue in build


JosephSalisbury:
```
5.25s$ go get -u github.com/golang/lint/golint
# cd .; git clone https://go.googlesource.com/tools /home/travis/gopath/src/golang.org/x/tools
Cloning into '/home/travis/gopath/src/golang.org/x/tools'...
remote: Daily bandwidth rate limit exceeded for IP 52.22.60.255.
fatal: protocol error: bad pack header
package golang.org/x/tools/go/gcimporter15: exit status 128
```

zeisss:
see #148 for fix

PR #150, opened 2016-03-21
Add git cookies to avoid rate limiting

JosephSalisbury:
Fixes https://github.com/giantswarm/inago/issues/147


JosephSalisbury:
We have to use a tarball because travis has issues encrypting multiple files, see: https://github.com/travis-ci/travis.rb/issues/239

zeisss:
LGTM

JosephSalisbury:
It shouldn't, the builds still fail due to the rate limiting.

PR #156, opened 2016-03-21
Update: We generate new sliceIDs in newReq, not Req. So we must retur…

zeisss:
…n the right thing


controller/update.go
```diff
@@ -132,7 +134,7 @@ func (c controller) runAddWorker(ctx context.Context, req Request, opts UpdateOp
 
 	time.Sleep(time.Duration(opts.ReadySecs) * time.Second)
 
-	return req, nil
+	return newReq, nil
```
zeisss, line 137: the fix

JosephSalisbury:
LGTM

PR #162, opened 2016-03-22
Add unit testing to update

JosephSalisbury:
Fixes https://github.com/giantswarm/inago/issues/131
- [x] Add tests for methods in `update.go`
- [x] Fix `go vet` error: `controller/update_test.go:99: arg test.errMatcher in printf call is a function value, not a function call`

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -854,3 +854,36 @@ func TestController_WaitForStatus_Timeout(t *testing.T) {
 	err := c.WaitForStatus(context.Background(), req, desired, closer)
 	Expect(IsWaitTimeoutReached(err)).To(BeTrue()) // Because WaitForStatus is 0 nothing should happen but directly return the error
 }
+
+// TestController_Update tests the Update method of the controller.
+func TestController_Update(t *testing.T) {
+	RegisterTestingT(t)
+
+	controller, fleetMock := givenController()
+	fleetMock.On("GetStatusWithMatcher", mock.AnythingOfType("func(string) bool")).Return(
+		[]fleet.UnitStatus{
+			{
+				Name: "test-main@1.service",
+			},
+		},
+		nil,
+	)
+
+	request := Request{
+		RequestConfig: RequestConfig{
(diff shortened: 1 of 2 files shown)


JosephSalisbury:
@giantswarm/cena RFR - I could spend ages on this, but I've got things to do, so I'm going to stop here.

@zeisss I couldn't hack that error issue, so I've removed the offending test. We'll see if it happens again.

Further, I've logged https://github.com/giantswarm/inago/issues/171 to further update testing.

Nesurion:
can we get more comments for the tests? I was thinking one short comment before each test case, explaining what the test case.

Nesurion:
LGTM

controller/controller_test.go
```diff
@@ -854,3 +854,36 @@ func TestController_WaitForStatus_Timeout(t *testing.T) {
...
+				Name: "test-main@1.service",
+			},
+		},
+		nil,
+	)
+
+	request := Request{
+		RequestConfig: RequestConfig{
+			Group:    "test",
+			SliceIDs: []string{"1"},
```
zeisss, line 875: Do SliceIDs make sense for an Update? I would expect their presence to raise an error.

controller/update_test.go
```diff
@@ -0,0 +1,267 @@
...
+	var tests = []struct {
+		function   func(context.Context, Request) (*task.Task, error)
+		ctx        context.Context
+		req        Request
+		errMatcher func(err error) bool
+	}{
+		// Test a task that does nothing.
+		{
+			function: func(ctx context.Context, req Request) (*task.Task, error) {
+				taskObject, _ := testController.TaskService.Create(
```
zeisss, line 55: why not just `return testController...`?

xh3b4sd:
Here are open questions. Do we need to address them?

zeisss:
The first one might be nice to be discussed.

JosephSalisbury:
Bah, I seem to have missed these. Uncertain.

PR #163, opened 2016-03-22
No err if update not needed

### LINKED ISSUE #143: inagoctl update throws error in case update is not needed
We might want to think about handling this case as a success?

Nesurion:
fixes #143

### DIFF
--- a/cli/common.go
+++ b/cli/common.go
@@ -91,7 +91,7 @@ func maybeBlockWithFeedback(ctx context.Context, bctx blockWithFeedbackCtx) {
 		}
 
 		if controller.IsUnitsAlreadyUpToDate(taskObject.Error) {
-			newLogger.Info(newCtx, taskObject.Error.Error())
+			newLogger.Info(ctx, "Not updating group '%s': (%s)", bctx.Request.Group, taskObject.Error.Error())
 			return
 		}
 
--- a/cli/common.go
+++ b/cli/common.go
@@ -90,13 +90,18 @@ func maybeBlockWithFeedback(ctx context.Context, bctx blockWithFeedbackCtx) {
 			os.Exit(1)
 		}
 
+		if controller.IsUnitsAlreadyUpToDate(taskObject.Error) {
+			newLogger.Info(newCtx, taskObject.Error.Error())
+			return
+		}
+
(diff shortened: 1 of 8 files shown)


Nesurion:
output look like this now:
`2016-03-22 13:43:47.896 | INFO     | context.Background: Not updating group 'mygroup': (units already up to date)`

Nesurion:
@giantswarm/cena RFR

cli/common.go
```diff
@@ -95,11 +95,16 @@ func maybeBlockWithFeedback(ctx context.Context, bctx blockWithFeedbackCtx) {
 			os.Exit(1)
 		}
 
+		if controller.IsUnitsAlreadyUpToDate(taskObject.Error) {
+			newLogger.Info(ctx, "Not updating group '%s': (%s)", bctx.Request.Group, taskObject.Error.Error())
```
zeisss: why the parantheses around the error text?

Nesurion: To match the other outputs. https://github.com/giantswarm/inago/blob/no-err-if-update-not-needed/cli/common.go#L105

zeisss: but there is no double colon there.

Nesurion: you are absolutely right! ;)

cli/common.go
```diff
@@ -128,7 +128,7 @@ func maybeBlockWithFeedback(ctx context.Context, bctx blockWithFeedbackCtx) {
 	} else {
 		newLogger.Info(
 			ctx,
-			"Succeeded to %s %d %v for group '%s': %v.",
+			"Succeeded to %s %d %v for group '%s'. %v.",
```
zeisss: leave double colon here please

cli/common.go
```diff
@@ -108,7 +108,7 @@ func maybeBlockWithFeedback(ctx context.Context, bctx blockWithFeedbackCtx) {
 			} else {
 				newLogger.Error(
 					ctx,
-					"Failed to %s %d %v for group '%s': %v. (%s)",
+					"Failed to %s %d %v for group '%s'. %v. (%s)",
```
zeisss: move the colon behind the parantheses please. (or are you writing sentences like this?)

JosephSalisbury:
LGTM

PR #169, opened 2016-03-22
Small improvements for the instance test

zeisss:
Just a grep to check the unit is actually active.

ping @JosephSalisbury


JosephSalisbury:
LGTM

PR #185, opened 2016-03-29
controller: Submit test now keeps track of submitted units and return…

zeisss:
…s a status for them.


Nesurion:
@giantswarm/cena RFR

zeisss:
LGTM

PR #194, opened 2016-03-30
Fix max-growth zero behaviour

### LINKED ISSUE #80: update: Fix behaviour of --max-growth=0
Might be the `--max-growth=0` flag is problematic. Having this setting given needs to remove slices first. It also needs to be validated if the settings are not destructive. When you have 2 slices running and `--min-alive=2` and `--max-growth=0` this will not work and should throw an error.

JosephSalisbury:
Fixes https://github.com/giantswarm/inago/issues/80
- [x] Stop infinite loop in general case
- [x] Work out how to fix the thing
- [x] Re-add goroutines in update
- [x] Fix Tim's notes
- [x] Add further tests for unit slices management

### DIFF
--- a/controller/update.go
+++ b/controller/update.go
@@ -107,10 +107,15 @@ func (c controller) isGroupAdditionAllowed(ctx context.Context, req Request, max
 }
 
 func (c controller) addFirst(ctx context.Context, req Request, opts UpdateOptions) error {
+	c.Config.Logger.Debug(ctx, "controller: running addFirst")
+
+	c.Config.Logger.Debug(ctx, "controller: running add worker")
 	newReq, err := c.runAddWorker(ctx, req, opts)
 	if err != nil {
 		return maskAny(err)
 	}
+
+	c.Config.Logger.Debug(ctx, "controller: checking number of running slices")
 	n, err := c.getNumRunningSlices(ctx, newReq)
 	if err != nil {
 		return maskAny(err)
@@ -118,6 +123,8 @@ func (c controller) addFirst(ctx context.Context, req Request, opts UpdateOption
 	if n != len(newReq.SliceIDs) {
 		return maskAnyf(updateFailedError, "addFirst: slice not running: %d != %v", n, newReq.SliceIDs)
 	}
+
+	c.Config.Logger.Debug(ctx, "controller: running remove worker")
 	err = c.runRemoveWorker(ctx, req)
 	if err != nil {
 		return maskAny(err)
@@ -151,15 +158,21 @@ func (c controller) runAddWorker(ctx context.Context, req Request, opts UpdateOp
 }
 
 func (c controller) removeFirst(ctx context.Context, req Request, opts UpdateOptions) error {
+	c.Config.Logger.Debug(ctx, "controller: running removeFirst")
+
+	c.Config.Logger.Debug(ctx, "controller: running remove worker")
 	err := c.runRemoveWorker(ctx, req)
 	if err != nil {
 		return maskAny(err)
 	}
 
+	c.Config.Logger.Debug(ctx, "controller: running add worker")
 	newReq, err := c.runAddWorker(ctx, req, opts)
 	if err != nil {
 		return maskAny(err)
 	}
+
+	c.Config.Logger.Debug(ctx, "controller: checking number of running slices")
 	n, err := c.getNumRunningSlices(ctx, newReq)
 	if err != nil {
 		return maskAny(err)
@@ -172,11 +185,13 @@ func (c controller) removeFirst(ctx context.Context, req Request, opts UpdateOpt
 }
 
 func (c controller) runRemoveWorker(ctx context.Context, req Request) error {
+	c.Config.Logger.Debug(ctx, "controller: executing stop action, req: %v", req)
 	// Stop.
 	if err := c.executeTaskAction(c.Stop, ctx, req); err != nil {
 		return maskAny(err)
 	}
 
+	c.Config.Logger.Debug(ctx, "controller: executing destroy action, req: %v", req)
 	// Destroy.
 	if err := c.executeTaskAction(c.Destroy, ctx, req); err != nil {
 		return maskAny(err)
@@ -252,7 +267,7 @@ func (c controller) UpdateWithStrategy(ctx context.Context, req Request, opts Up
 		for {
 			changeAttemptMade := false
 
-			c.Config.Logger.Debug(ctx, "controller: adding slice: %v", sliceID)
+			c.Config.Logger.Debug(ctx, "controller: attempting to add slice: %v", sliceID)
 			// add
 			maxGrowth := opts.MaxGrowth + numTotal - opts.MinAlive - int(addInProgress)
 			ok, err := c.isGroupAdditionAllowed(ctx, req, maxGrowth)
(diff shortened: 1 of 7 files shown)


JosephSalisbury:
As far as I can tell, the test added in https://github.com/giantswarm/inago/pull/194/commits/a7ca6bf96e090f0002f07b4dc492f4aa2538f890 should actually _fail_, instead of passing - the idea was to put in a regression test against the behaviour seen in https://github.com/giantswarm/inago/issues/80#issuecomment-202934288.

Still investigating why the test passes - my reckoning is currently on either a mismatch between arguments to `UpdateWithStrategy`, or an issue with the dummy fleet implementation.

JosephSalisbury:
Okay, I've gotten everything green, so I'm fairly certain this is fixed. I want to do a bit more testing / refactoring/cleanup on Monday, then i'll properly submit it for review.

With regards to root cause, it looks like `isGroupAdditionAllowed` and `isGroupRemovalAllowed` didn't take into account the new slice IDs that were created. This meant that in a max-growth zero situation, we removed one slice, and then were attempting to see if we could add (which we can't because max growth), or remove (which we couldn't because there's only one slice). This locked up the entire update loop.

This changeset tracks the current slices in use, and uses that for the addition/removal calculations.

For the interested, the reason the initial unit test didn't correctly test the situation was due to the lack of `MachineStatus` - `inago` didn't think the slices were running at all. Adding the `MachineStatus` fields to the dummy fleet implementation helped this.

I've also removed the goroutines around the addFirst and removeFirst calls, I don't get why they were there, and they caused me headaches. Happy to add them back if they were there for _reasons_

controller/update.go
```diff
@@ -73,49 +73,64 @@ func (c controller) getNumRunningSlices(ctx context.Context, req Request) (int,
...
 
-func (c controller) addFirst(ctx context.Context, req Request, opts UpdateOptions) error {
+func (c controller) addFirst(ctx context.Context, req Request, opts UpdateOptions) ([]string, error) {
+	c.Config.Logger.Debug(ctx, "controller: running addFirst")
+
+	c.Config.Logger.Debug(ctx, "controller: running add worker")
 	newReq, err := c.runAddWorker(ctx, req, opts)
 	if err != nil {
-		return maskAny(err)
+		return []string{}, maskAny(err)
```
xh3b4sd: You can simply return `nil` here.

controller/update.go
```diff
@@ -209,6 +232,30 @@ type UpdateOptions struct {
 	ReadySecs int
 }
 
+// updateCurrentSliceIDs updates the list of current slice IDs,
+// removing the slice that was modified, and adding any new slice IDs.
+func (c controller) updateCurrentSliceIDs(ctx context.Context, currentSliceIDs []string, modifiedSliceIDs []string, newSliceIDs []string) []string {
```
xh3b4sd, line 238: You can ease the argument list as follows.

```
updateCurrentSliceIDs(ctx context.Context, currentSliceIDs, modifiedSliceIDs, newSliceIDs []string)
```

controller/update.go
```diff
@@ -232,53 +281,93 @@ func (c controller) UpdateWithStrategy(ctx context.Context, req Request, opts Up
...
 				break
 			}
 
+			c.Config.Logger.Debug(ctx, "controller: finished attempts to make changes")
+			if !changeAttemptMade {
+				c.Config.Logger.Warning(ctx, "controller: failed to make any changes")
+				numFailedChangeAttempts++
+			}
+			// If we've failed too many times, just give up completely :(
+			if numFailedChangeAttempts > c.Config.MaxFailedChangeAttempts {
```
xh3b4sd: IMO we should fail as soon as the max is reached. Thus `>=`.

controller/update_test.go
```diff
@@ -207,6 +212,38 @@ func TestIsGroupRemovalAllowed(t *testing.T) {
...
+		modifiedSliceIDs []string
+		newSliceIDs      []string
+		output           []string
+	}{
+		{
+			currentSliceIDs:  []string{"lol", "kek"},
+			modifiedSliceIDs: []string{"lol"},
+			newSliceIDs:      []string{"02b"},
+			output:           []string{"kek", "02b"},
+		},
```
xh3b4sd, line 257: IMO some combinations are missing. Lets add some tests having more slice IDs and even none slice IDs. That way we can make sure more "complicated" cases are working and the code does not panic when lists are empty.

controller/update.go
```diff
@@ -209,6 +232,30 @@ type UpdateOptions struct {
...
+		}
+	}
+	currentSliceIDs = append(currentSliceIDs, newSliceIDs...)
+
+	c.Config.Logger.Debug(ctx, "updated slice IDs: %v", currentSliceIDs)
+
+	return currentSliceIDs
+}
+
 func (c controller) UpdateWithStrategy(ctx context.Context, req Request, opts UpdateOptions) error {
```
xh3b4sd, line 263: I would like to consider this complain as well: https://goreportcard.com/report/github.com/giantswarm/inago#gocyclo.

JosephSalisbury, line 263: Yeah, it would be cool to improve this, but I'd rather it happen in a separate PR.

JosephSalisbury:
I've removed the locking check, I couldn't get it into a state where I was happy with it, and I'm unconvinced it will happen again. If it does, we can look back into it.

JosephSalisbury:
It also keeps the scope of this PR smaller, that is, fixing the issue where slice ids weren't handled correctly for max-growth zero

controller/update.go
xh3b4sd, line 263: This is only about splitting up one function into multiple functions to prevent deep nesting, but ok. Do it on your behalf.

JosephSalisbury:
@giantswarm/cena RFR pretty plz

controller/update_test.go
```diff
@@ -34,8 +35,8 @@ func getTestController() (controller, *fleet.DummyFleet) {
 	newControllerConfig.TaskService = newTaskService
 	newControllerConfig.Logger = newLogger
 	newControllerConfig.WaitCount = 1
-	newControllerConfig.WaitSleep = 1 * time.Millisecond
-	newControllerConfig.WaitTimeout = 30 * time.Second
+	newControllerConfig.WaitSleep = 300 * time.Millisecond
+	newControllerConfig.WaitTimeout = 5 * time.Second
```
xh3b4sd, line 39: Why that change?

controller/update_test.go
xh3b4sd, line 257: Looks good now.

fabfile.py
```diff
@@ -11,7 +13,7 @@
 env.hosts = ['core@ec2-52-58-89-242.eu-central-1.compute.amazonaws.com']
 env.disable_known_hosts = True
 env.colorize_errors = True
-env.command_timeout = 60 * 5
+env.command_timeout = 60 * 10
```
xh3b4sd, line 16: Why that change?

controller/update_test.go
JosephSalisbury, line 39: So if it is locked up, you don't have to wait as long during testing.

xh3b4sd:
Overall LGTM. Only minor questions.

xh3b4sd:
One more question. This "max attempts failed" thingy is now gone. Why is that? Why did we want that?

fabfile.py
JosephSalisbury, line 16: This controls how long fabric will wait before it timesout, so integration tests can only take as long as this (in seconds) - the new integration test ups the run time to ~6 minutes, so I increased it, as well as adding a little bit of breathing room.

JosephSalisbury:
In the failure case where we can neither add nor remove groups - which we had here, because of the slice IDs management issue - the update will lock up completely, and goe off in an infinite loop. The max attempts safety would notice this, and error out.

I removed it because the fix in this PR should stop that happening, and I wasn't convinced that the effort/complexity of keeping it - after re-adding the goroutines - was worth it. If a similar situation happens again, we should probably revisit the design, instead of adding a safety.

controller/update_test.go
JosephSalisbury, line 39: Largely arbitrary, but I found it helped :D

Issue #103, opened 2016-03-15
Add up command


zeisss:
https://gigantic.slack.com/archives/team-cena/p1456481207000400

add inagoctl up to submit + start units
inagoctl start should NOT create units, if they don't exist

PR #195, opened 2016-03-31
Add up command

JosephSalisbury:
Fixes https://github.com/giantswarm/inago/issues/103

### DIFF
--- a/cli/up.go
+++ b/cli/up.go
@@ -1,0 +1,26 @@
+package cli
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var (
+	upCmd = &cobra.Command{
+		Use:   "up <group> [scale]",
+		Short: "Bring a group up",
+		Long:  "Submit a group, with an optional scale, and start it",
+		Run:   upRun,
+	}
+)
+
+func upRun(cmd *cobra.Command, args []string) {
+	submitRun(cmd, args)
+
+	// If a scale argument has been passed to submit,
+	// remove it from the args list, as start doesn't want it.
+	if len(args) > 1 {
+		args = args[:1]
+	}
+
+	startRun(cmd, args)
+}
--- a/cli/inagoctl.go
+++ b/cli/inagoctl.go
@@ -87,6 +87,7 @@ func init() {
 	MainCmd.AddCommand(startCmd)
 	MainCmd.AddCommand(stopCmd)
 	MainCmd.AddCommand(destroyCmd)
+	MainCmd.AddCommand(upCmd)
 	MainCmd.AddCommand(updateCmd)
 	MainCmd.AddCommand(validateCmd)
 	MainCmd.AddCommand(versionCmd)
--- a/int-tests/000-usage.t
+++ b/int-tests/000-usage.t
@@ -11,6 +11,7 @@
     start       Start a group
     stop        Stop a group
     destroy     Destroy a group
+    up          Bring a group up
     update      Update a group
     validate    Validate groups
     version     Print version


JosephSalisbury:
@giantswarm/cena RFR

cli/up.go
```diff
@@ -0,0 +1,26 @@
...
+	upCmd = &cobra.Command{
+		Use:   "up <group> [scale]",
+		Short: "Bring a group up",
+		Long:  "Submit a group, with an optional scale, and start it",
+		Run:   upRun,
+	}
+)
+
+func upRun(cmd *cobra.Command, args []string) {
+	submitRun(cmd, args)
```
zeisss, line 17: if submit raises an error, do we still continue here?

zeisss, line 17: how does this behave for already existing slices?

JosephSalisbury, line 17: error during submit:

```
core@core-01 ~ $ ls -R
.:
example  inagoctl

./example:
example-unit.service
core@core-01 ~ $ cat ./example/example-unit.service
[Unit]
Description=Inago Demo Unit

[Service]
ExecStart=/bin/bash -c "while true; do echo Hi; sleep 10; done"
core@core-01 ~ $ ./inagoctl up example 4
2016-03-31 10:22:23.337 | ERROR    | context.Background: [{/usr/code/.gobuild/src/github.com/giantswarm/inago/cli/submit.go:47: } {/usr/code/.gobuild/src/github.com/giantswarm/inago/cli/submit.go:80: invalid scale: must be 1 for unscalable groups}]
core@core-01 ~ $ ./inagoctl status example
2016-03-31 10:22:29.651 | ERROR    | context.Background: Failed to find group 'example'.
```

so, if submit errors and exits - which it seems to do on all errors, we're fine

for already existing slices:

```
core@core-01 ~ $ ls -R
.:
example  inagoctl

./example:
example-unit@.service
core@core-01 ~ $ cat ./example/example-unit\@.service
[Unit]
Description=Inago Demo Unit

[Service]
ExecStart=/bin/bash -c "while true; do echo Hi; sleep 10; done"
core@core-01 ~ $ ./inagoctl submit example
2016-03-31 10:20:55.241 | INFO     | context.Background: Succeeded to submit group 'example'.
core@core-01 ~ $ ./inagoctl status example
Group        Units  FDState  FCState  SAState   IP            Machine

example@b5a  *      loaded   loaded   inactive  172.17.8.103  2eb8b3bb62894c269df58a00248a68eb

core@core-01 ~ $ ./inagoctl up example
2016-03-31 10:21:09.618 | INFO     | context.Background: Succeeded to submit group 'example'.
2016-03-31 10:21:14.625 | INFO     | context.Background: Succeeded to start 2 slices for group 'example': [899 b5a].
core@core-01 ~ $ ./inagoctl status example
Group        Units  FDState   FCState   SAState  IP            Machine

example@899  *      launched  launched  active   172.17.8.102  49e7e432625f4d9fa3e5071cf3f9c26e
example@b5a  *      launched  launched  active   172.17.8.103  2eb8b3bb62894c269df58a00248a68eb
```

uncertain about the behaviour here - it is logical (it's the same as two submits, and then a start), and i'm not certain how we'd get around it anyway.

zeisss, line 17: well, I think the expectation of `inagoctl up mygroup 4` would be that afterwards there are 4 slices. We probably can involve @marians here. 

What does this mean it there are already 2 slices? Do we just want to submit 2 slices and then start all 4?

JosephSalisbury, line 17: If there are already 2 slices submitted of 'mygroup', and you up 2 slices of 'mygroup', then you will end up with 4 running slices, as up == submit then start, meaning that the 2 previously submitted slices will also get started.

I think having starting all four slices is logically consistent. If we see that people find this confusing, we could iterate on it - for example, up could grumble if there is an existing group.

marians, line 17: Sounds good to me. I'm not aware of any global UX heuristic that could help us here. We have to decide and be consistent regarding this: Do we want (a) arguments like "number of slices to deal with" to affect the resulting number (add so many that we have _n_ in the end)? Or (b) should they set the number for the current command (add _n_)?

(b) is simpler, more straightforward in my opinion and also a simpler promise to keep in a distributed team.

zeisss, line 17: > If there are already 2 slices submitted of 'mygroup', and you up 2 slices of 'mygroup', then you will end up with 4 running slices, as up == submit then start, meaning that the 2 previously submitted slices will also get started.

this sounds like the current implementation, but what would we expect "up" to do? If I do `inago up mygroup 4` and two are already there, do I end up with 6 or 4?

JosephSalisbury, line 17: I think (b) is more in line with the current design, as it is now.

JosephSalisbury, line 17: @zeisss sorry, i misunderstood, that is the current implementation.

I think if I `up`'d 4, I would expect four to be there, but I don't think it's surprising that if you submit 2 of the same group beforehand, they also get started.

zeisss, line 17: ack

zeisss, line 17: ack, I also would expect: "up 4" makes the end number 4, all started.

JosephSalisbury, line 17: so, if there were already 2 submitted units, you'd expect 2 units to be started, and then all 4 started?

i don't feel like this is consistent with the current behaviour of submit - submit 3 will _add_ 3, while it sounds like up 3 will _make sure there are 3_

zeisss, line 17: ack, but thats just my expectation. thats why should talk with @marians

JosephSalisbury, line 17: if i understand https://github.com/giantswarm/inago/pull/195#discussion_r58044528 correctly, the current implementation sounds good to @marians, assuming we are consistent with it (in any later commands)

zeisss, line 17: k :)

zeisss:
LGTM

PR #200, opened 2016-04-04
adds ssh tunneling

### LINKED ISSUE #174: Support more than http api or local domain socket
For our production servers, we are not exposing fleet over http, not even within the VPN. So to use inago there, we would need to copy it over along with the unit files to one of the servers and execute it remotely. 

These are unnecessary steps. OpenSSH ships domain socket proxying with [6.7](https://lwn.net/Articles/609321/), but thats not in the standard version of OS X. Any other hints how to use this more easily?

Nesurion:
This PR adds a `--tunnel` flag, which enables the user to use an ssh tunnel to connect to fleet.
The implementation is based heavily on https://github.com/coreos/fleet/blob/master/fleetctl/fleetctl.go and uses the ssh client in https://github.com/coreos/fleet/blob/master/fleetctl/ssh.go.

fixes #174

### DIFF
--- a/cli/inagoctl.go
+++ b/cli/inagoctl.go
@@ -4,6 +4,7 @@ package cli
 
 import (
 	"net/url"
+	"time"
 
 	"github.com/spf13/cobra"
 	"golang.org/x/net/context"
@@ -21,6 +22,12 @@ var (
 		FleetEndpoint string
 		NoBlock       bool
 		Verbose       bool
+
+		Tunnel                string
+		SSHUsername           string
+		SSHTimeout            time.Duration
+		StrictHostKeyChecking bool
+		KnownHostsFile        string
 	}
 
 	fs             filesystemspec.FileSystem
@@ -56,6 +63,14 @@ var (
 			newFleetConfig := fleet.DefaultConfig()
 			newFleetConfig.Endpoint = *URL
 			newFleetConfig.Logger = newLogger
+			newFleetConfig.Tunnel = fleet.SSHTunnel{
+				Tunnel:                globalFlags.Tunnel,
+				Username:              globalFlags.SSHUsername,
+				Timeout:               globalFlags.SSHTimeout,
+				StrictHostKeyChecking: globalFlags.StrictHostKeyChecking,
+				KnownHostsFile:        globalFlags.KnownHostsFile,
+			}
+
 			newFleet, err = fleet.NewFleet(newFleetConfig)
 			if err != nil {
 				panic(err)
@@ -81,6 +96,12 @@ func init() {
 	MainCmd.PersistentFlags().StringVar(&globalFlags.FleetEndpoint, "fleet-endpoint", "unix:///var/run/fleet.sock", "endpoint used to connect to fleet")
 	MainCmd.PersistentFlags().BoolVar(&globalFlags.NoBlock, "no-block", false, "block on synchronous actions")
 	MainCmd.PersistentFlags().BoolVarP(&globalFlags.Verbose, "verbose", "v", false, "verbose output")
+
+	MainCmd.PersistentFlags().StringVar(&globalFlags.Tunnel, "tunnel", "", "use a tunnel to communicate with fleet")
+	MainCmd.PersistentFlags().StringVar(&globalFlags.SSHUsername, "ssh-username", "core", "username to use when connecting to CoreOS machine")
+	MainCmd.PersistentFlags().DurationVar(&globalFlags.SSHTimeout, "ssh-timeout", time.Duration(10*time.Second), "timeout in seconds when establishing the connection via SSH")
+	MainCmd.PersistentFlags().BoolVar(&globalFlags.StrictHostKeyChecking, "strict-host-key-checking", true, "verify host keys presented by remote machines before initiating SSH connections")
+	MainCmd.PersistentFlags().StringVar(&globalFlags.KnownHostsFile, "known-hosts-file", "~/.fleetctl/known_hosts", "file used to store remote machine fingerprints")
 
 	MainCmd.AddCommand(submitCmd)
 	MainCmd.AddCommand(statusCmd)
--- a/cli/inagoctl.go
+++ b/cli/inagoctl.go
@@ -23,11 +23,11 @@ var (
 		NoBlock       bool
 		Verbose       bool
 
-		Tunnel                string
-		SSHUsername           string
-		SSHTimeout            time.Duration
-		StrictHostKeyChecking bool
-		KnownHostsFile        string
+		Tunnel                   string
+		SSHUsername              string
+		SSHTimeout               time.Duration
+		SSHStrictHostKeyChecking bool
+		SSHKnownHostsFile        string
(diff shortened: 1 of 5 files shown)


Nesurion:
@giantswarm/cena RFR

cli/inagoctl.go
```diff
@@ -82,6 +97,12 @@ func init() {
 	MainCmd.PersistentFlags().BoolVar(&globalFlags.NoBlock, "no-block", false, "block on synchronous actions")
 	MainCmd.PersistentFlags().BoolVarP(&globalFlags.Verbose, "verbose", "v", false, "verbose output")
 
+	MainCmd.PersistentFlags().StringVar(&globalFlags.Tunnel, "tunnel", "", "use a tunnel to communicate with fleet")
+	MainCmd.PersistentFlags().StringVar(&globalFlags.SSHUsername, "ssh-username", "core", "username to use when connecting to CoreOS machine")
+	MainCmd.PersistentFlags().DurationVar(&globalFlags.SSHTimeout, "ssh-timeout", time.Duration(10*time.Second), "timeout in seconds when establishing the connection via SSH")
+	MainCmd.PersistentFlags().BoolVar(&globalFlags.StrictHostKeyChecking, "strict-host-key-checking", true, "verify host keys presented by remote machines before initiating SSH connections")
+	MainCmd.PersistentFlags().StringVar(&globalFlags.KnownHostsFile, "known-hosts-file", "~/.fleetctl/known_hosts", "file used to store remote machine fingerprints")
```
xh3b4sd: What about consistently prefixing the flags with `ssh`?

fleet/fleet.go
```diff
@@ -131,6 +132,15 @@ type Fleet interface {
 func NewFleet(config Config) (Fleet, error) {
 	var trans http.RoundTripper
 
+	tunneling := config.Tunnel.Tunnel != ""
+	// If a tunnel in provided we need to overwrite the http.Transport.Dial function
```
xh3b4sd: `in` -> `is`

fleet/fleet.go
```diff
@@ -131,6 +132,15 @@ type Fleet interface {
...
+	// to use the tunnel
+	if tunneling {
+		trans = &http.Transport{
+			Dial: config.Tunnel.NewDialFunc(config),
+		}
+	}
```
xh3b4sd: This looks like the transport is going to be overwritten in the upcoming switch, in case `config.Endpoint.Scheme` is `unix` or `file`.

cli/inagoctl.go
Nesurion: True, but I'd like `tunnel` to be just called `tunnel`, as it is shorter and fleetctl also just calles it tunnel.

fleet/ssh.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	tunnelFunc := sshClient.Dial
+
+	if config.Endpoint.Scheme == "unix" || config.Endpoint.Scheme == "file" {
+		tgt := config.Endpoint.Path
+		tunnelFunc = func(string, string) (net.Conn, error) {
+			cmd := fmt.Sprintf(`fleetctl fd-forward %s`, tgt)
+			return ssh.DialCommand(sshClient, cmd)
+		}
+	}
```
xh3b4sd, line 37: 1. I don't get what `tgt` means. 
2. I don't get the stuff that is being done here. Can you explain why we need this?

Nesurion, line 37: 1. `tgt` = target, but I'm with you on that. I'll rename it to `target`
2. The best thing I can give you is: https://github.com/coreos/fleet/blob/master/fleetctl/fdforward.go
   I can't provide a detailed explanation.

fleet/fleet.go
Nesurion: https://github.com/giantswarm/inago/blob/ssh-tunnel/fleet/fleet.go#L167
the if statement in the `unix` case makes sure that the dail function is not overwritten in case tunneling is enabled.

xh3b4sd: I would say we don't need to execute the whole switch logic in case we tunnel through ssh. Then line 167 can be dropped and the code flow is more straight forward. What do you think?

fleet/ssh.go
xh3b4sd, line 37: It looks like the tunneling is actually done through a ssh remote command being executed. stdin and stdout are then proxied between the client and the server side.

xh3b4sd, line 37: Here the remote command is `fleetctl fd-forward <target>`.

fleet/fleet.go
```diff
@@ -131,18 +132,15 @@ type Fleet interface {
...
-	case "unix", "file":
-		if config.Endpoint.Host != "" {
-			// This commonly happens if the user misses the leading slash after the
-			// scheme. For example, "unix://var/run/fleet.sock" would be parsed as
-			// host "var".
-			return nil, maskAnyf(invalidEndpointError, "cannot connect to host %q with scheme %q", config.Endpoint.Host, config.Endpoint.Scheme)
+	tunneling := config.Tunnel.Tunnel != ""
+	// If a tunnel is provided we need to overwrite the http.Transport.Dial function
+	// to use the tunnel
+	if tunneling {
```
Nesurion, line 138: @xh3b4sd not sure if this is better :/ I dislike the code replication with 156 ff. What do you think?

xh3b4sd, line 138: Why do you dislike it?

xh3b4sd:
Another thing. Can we test this somehow? When we already have integration tests we probably can add one for this feature right?

fleet/fleet.go
Nesurion, line 138: no matter if the tunnel flag is set or not: if the endpoint is unix socket then we need to make some changes (https://github.com/giantswarm/inago/blob/ssh-tunnel/fleet/fleet.go#L161-L164) to the endpoint in the config.
Since this is always true for socket connections I feel we have two places in the code that do the same thing now

xh3b4sd, line 138: To mee this is about two different things. You only changed the codeflow here. For me this is now more readable and better understandable because now it is linear and straight forward. 

Towards the other point, you are right. How to fix this? I guess it is what it is. Two places is not that bad. In go you repeat yourself quite often. This is sometimes a desired design decisions based on the golang devs intuitions.

xh3b4sd, line 138: To mee this is about two different things. You only changed the codeflow here. For me this is now more readable and better understandable because now it is linear and straight forward. 

Towards the other point, you are right. How to fix this? I guess it is what it is. Two places is not that bad. In go you repeat yourself quite often. This is sometimes a desired design decisions based on the golang devs intuitions.

xh3b4sd:
@Nesurion and I discussed several things. To me it is now a little bit more clear how our integration test setup works. We will add an integration test for the tunneling. I don't care if now in this PR or separately. The codeflow conditionals are good enough as they are. Lets go with this. 

LGTM

Nesurion:
int tests will be handled in a separate PR.

PR #211, opened 2016-04-06
ignore failed state when stopping units

### LINKED ISSUE #206: On updating, we should ignore failed units
On updating, when we remove units, we should ignore units with a failed state - they're going away anyway.

xh3b4sd:
Fixes #206 

RFR @JosephSalisbury

### DIFF
--- a/controller/status.go
+++ b/controller/status.go
@@ -309,6 +309,9 @@ func (a Aggregator) AggregateStatus(fc, fd, sa, ss string) (Status, error) {
 	return aggregatedStatuses[0], nil
 }
 
+// UnitHasStatus determines if a given unit's status is effectivly equal to a
+// set of given statuses. This method provides status mapping of
+// AggregateStatus and compares the result with the given set of statuses.
 func (a Aggregator) UnitHasStatus(us fleet.UnitStatus, statuses ...Status) (bool, error) {
 	if len(statuses) == 0 {
 		return false, maskAny(invalidArgumentError)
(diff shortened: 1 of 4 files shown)


controller/status.go
```diff
@@ -309,19 +309,25 @@ func (a Aggregator) AggregateStatus(fc, fd, sa, ss string) (Status, error) {
 	return aggregatedStatuses[0], nil
 }
 
-func (a Aggregator) unitHasStatus(us fleet.UnitStatus, status Status) (bool, error) {
+func (a Aggregator) UnitHasStatus(us fleet.UnitStatus, statuses ...Status) (bool, error) {
```
JosephSalisbury, line 315: This could use a godoc

controller/update.go
```diff
@@ -300,11 +300,16 @@ func (c controller) UpdateWithStrategy(ctx context.Context, req Request, opts Up
...
-			// add
-			maxGrowth := opts.MaxGrowth + numTotal - opts.MinAlive - int(addInProgress)
+
+			currentSliceReq := req
+			// Calculating the number of groups allowed to be added during this
+			// iteration respects the total number of running groups. This is because
+			// we later going to check how many are actually running after the
+			// addition. So the next iteration will calculate and check again and
+			// respect the total number of running groups. See also
+			// isGroupAdditionAllowed.
```
JosephSalisbury, line 320: I wouldn't mind seeing a similar explanation later, for the `minAlive` calculation - particularly why the two are different. This caught us out earlier.

JosephSalisbury:
LGTM, few comments

controller/update.go
xh3b4sd, line 320: The calculation for `minAlive` is more straight forward. I would keep it as it is. There was no problem. I would not know what to document there besides the rather obvious information.

JosephSalisbury, line 320: Works for me <3

Issue #192, opened 2016-03-30
Example: Elasticsearch


zeisss:
We want to determine whether the current inago feature set is sufficient for setting up a production-ish ELK logging set up. If it is, we should work on the example; if it isn't, we need to determine what features are missing. This issue tracks the work of determining that.
- Elasticsearch
- Kibana
- Logstash-Forwarder (global)

zeisss:
Look at what releaseit does here.

xh3b4sd:
I looked at the unit files created by releaseit. My main concern was that we do only support `.service` unit types. 

This is my releaseit command to check the unit files.

```
releaseit create monitoring --group=xxx --dry-run --dry-run-path=tmp --env=stage/prod --stack=foo --provider aws
```

My thoughts.
- Depending on the used storage we are fine. AFAIK when using local storage we need `.mount` unit files. 
- Releaseit also creates the `elasticsearch-curator`, which uses a `.timer` unit to continuously cleanup old data. 
- We may could ignore that particular fact for a minimal example. AFAIK it is possible to handle global units with Inago. 

Anyway, I don't think the fix to support all kinds of unit types would be that expensive. Overall I think it should be possible to provide an ES example with little improvements in Inago.

zeisss:
@xh3b4sd FYI: Releaseit now has an `export-inago-path` flag in the `./releaseit create` command.

zeisss:
ping @giantswarm/cena

JosephSalisbury:
current todo:
- [ ] Sort out the elasticsearch group
- [ ] Write an example for it

JosephSalisbury:
@xh3b4sd i've stolen assignment :D

PR #214, opened 2016-04-06
Elasticsearch example

JosephSalisbury:
Fixes https://github.com/giantswarm/inago/issues/192

### DIFF
--- a/docs/elasticsearch.md
+++ b/docs/elasticsearch.md
@@ -1,0 +1,204 @@
+# Elasticsearch
+
+With Inago, it's a piece of cake to deploy an Elasticsearch cluster,
+as well as updating it without any downtime or data loss.
+
+It should be noted that this is a demo, and there are dangers to using it in a production environment without further testing.
+Saying that, if you want to do so, and have questions, the team is always available on IRC (`#giantswarm` on `freenode`) <3
+
+## Setup
+- We're using a CoreOS cluster running locally, with Vagrant. See https://coreos.com/os/docs/latest/booting-on-vagrant.html for further information.
+- There are three nodes in the cluster, and each node has 2048MB of memory allocated to it.
+- The `inago` binary, as well as the `elasticsearch` directory from `example`, have been copied to all the nodes in the cluster.
+
+## Starting the Elasticsearch cluster
+We're going to use Inago's `up` command to bring up multiple group slices of the `elasticsearch` group. `up` is equivalent to `submit`, followed by `start`.
+
+Due to the `Conflicts` statement in the unit file for the group, each group slice will be automatically scheduled onto separate machines.
+```
+core@core-01 ~ $ ./inagoctl up elasticsearch 3
+2016-04-06 16:57:39.644 | INFO     | context.Background: Succeeded to submit group 'elasticsearch'.
+2016-04-06 16:57:46.656 | INFO     | context.Background: Succeeded to start 3 slices for group 'elasticsearch': [1f4 8f1 9e9].
+```
+As you can see, the `elasticsearch` group has been submitted and started. Each group slice has an ID, which can be seen in the output.
+
+We can use the `status` command to inspect the state of the group. The _verbose_ flag is used so that the hash of the unit file is also printed. This will be interesting later, after we upgrade Elasticsearch. You can also see the the group slices have been scheduled onto multiple machines.
+```
+core@core-01 ~ $ ./inagoctl status -v elasticsearch
+Group              Units                      FDState   FCState   SAState  Hash                                      IP            Machine
+
+elasticsearch@2cf  elasticsearch@2cf.service  launched  launched  active   07969bf8caa1a7f8c73b35bb826229425d3264f6  172.17.8.103  1564e0d3c4ab4aadbe76b24cea90c62f
+elasticsearch@445  elasticsearch@445.service  launched  launched  active   07969bf8caa1a7f8c73b35bb826229425d3264f6  172.17.8.102  2dbbe125e237410bb946e7b0f6b95e4c
+elasticsearch@891  elasticsearch@891.service  launched  launched  active   07969bf8caa1a7f8c73b35bb826229425d3264f6  172.17.8.101  b89277d82f6a425eb83abb39c00485bc
+```
+
+We can also check Elasticsearch via its own API.
+```
+core@core-02 ~ $ curl -XGET 'http://172.17.8.101:9200/_cluster/health?pretty=true'
+{
+  "cluster_name" : "elk",
+  "status" : "green",
+  "timed_out" : false,
+  "number_of_nodes" : 3,
+  "number_of_data_nodes" : 3,
+  "active_primary_shards" : 0,
+  "active_shards" : 0,
+  "relocating_shards" : 0,
+  "initializing_shards" : 0,
+  "unassigned_shards" : 0,
+  "delayed_unassigned_shards" : 0,
+  "number_of_pending_tasks" : 0,
+  "number_of_in_flight_fetch" : 0,
+  "task_max_waiting_in_queue_millis" : 0,
+  "active_shards_percent_as_number" : 100.0
+}
+```
+The cluster status is `green`, and we have three nodes running. Hurrah!
+
+We'll add an index, and a document to that index, too.
+```
+core@core-02 ~ $ curl -XPUT '172.17.8.102:9200/inago-example-test-index?pretty'
+{
+  "acknowledged" : true
+}
+core@core-02 ~ $ curl -XPUT 'http://172.17.8.103:9200/inago-example-test-index/external/1?pretty' -d '{"nanana": "batman!"}'
+{
+  "_index" : "inago-example-test-index",
+  "_type" : "external",
+  "_id" : "1",
+  "_version" : 1,
+  "_shards" : {
+    "total" : 2,
+    "successful" : 2,
+    "failed" : 0
+  },
+  "created" : true
+}
+core@core-02 ~ $ curl -XGET '172.17.8.101:9200/inago-example-test-index/external/1?pretty'
+{
+  "_index" : "inago-example-test-index",
+  "_type" : "external",
+  "_id" : "1",
+  "_version" : 1,
+  "found" : true,
+  "_source" : {
+    "nanana" : "batman!"
+  }
+}
+```
+
+## Updating the Elasticsearch cluster
+Inago operates on _groups_ of unit files. If we look at the `elasticsearch@.service` file, in the `elasticsearch` directory, you can see the unit file for the group we're using in this example.
+
+The Docker image for this unit is set via the line:
+```
+Environment="IMAGE=elasticsearch:2.2"
+```
+
+The currently running version of Elasticsearch can be verified via the Elasticsearch API.
+```
+core@core-02 ~ $ curl 172.17.8.101:9200
+{
+  "name" : "Crime-Buster",
+  "cluster_name" : "inago-example",
+  "version" : {
+    "number" : "2.2.2",
+    "build_hash" : "fcc01dd81f4de6b2852888450ce5a56436fd5852",
+    "build_timestamp" : "2016-03-29T08:49:35Z",
+    "build_snapshot" : false,
+    "lucene_version" : "5.4.1"
+  },
+  "tagline" : "You Know, for Search"
+}
+```
+
+
+However, this isn't the latest version of Elasticsearch! We want to upgrade it to 2.3.
+
+We need to modify the line to read:
+```
+Environment="IMAGE=elasticsearch:2.3"
+```
+This will update us to use the later version of the Elasticsearch Docker image.
+
+Next, we're going to use the `update` command from Inago to perform the update of the Elasticsearch cluster.
+```
+core@core-01 ~ $ ./inagoctl update elasticsearch --max-growth=1 --min-alive=2 --ready-secs=60
+2016-04-06 17:11:54.026 | INFO     | context.Background: Succeeded to update 3 slices for group 'elasticsearch': [184 372 88f].
+```
+The arguments used here mean that Inago is allowed to create one additional slice during the update (meaning there will be no more than four instances of Elasticsearch running at any time), and that at least two instances of Elasticsearch have to be running at any time.
+
+The `ready-secs` flag determines how long to wait between rounds of updating. This value gives Elasticsearch enough time to replicate any necessary data.
+
+The IDs of the new group slices created during the update are printed in the output.
+
+We can check the Elasticsearch cluster health via the Elasticsearch API.
+```
+core@core-02 ~ $ curl -XGET 'http://172.17.8.101:9200/_cluster/health?pretty=true'
+{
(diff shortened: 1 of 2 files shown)


JosephSalisbury:
@giantswarm/cena @puja108 your thoughts, please and thank you <3

docs/elasticsearch.md
```diff
@@ -0,0 +1,204 @@
...
+With Inago, it's a piece of cake to deploy an Elasticsearch cluster,
+as well as updating it without any downtime or data loss.
+
+It should be noted that this is a demo, and there are dangers to using it in a production environment without further testing.
+Saying that, if you want to do so, and have questions, the team is always available on IRC (`#giantswarm` on `freenode`) <3
+
+## Setup
+- We're using a CoreOS cluster running locally, with Vagrant. See https://coreos.com/os/docs/latest/booting-on-vagrant.html for further information.
+- There are three nodes in the cluster, and each node has 2048MB of memory allocated to it.
+- The `inago` binary, as well as the `elasticsearch` directory from `example`, have been copied to all the nodes in the cluster.
```
puja108: we should add a general "installation" part to the getting started: https://github.com/giantswarm/inago/blob/master/docs/getting_started.md

docs/elasticsearch.md
```diff
@@ -0,0 +1,204 @@
+# Elasticsearch
+
+With Inago, it's a piece of cake to deploy an Elasticsearch cluster,
+as well as updating it without any downtime or data loss.
+
+It should be noted that this is a demo, and there are dangers to using it in a production environment without further testing.
+Saying that, if you want to do so, and have questions, the team is always available on IRC (`#giantswarm` on `freenode`) <3
+
+## Setup
+- We're using a CoreOS cluster running locally, with Vagrant. See https://coreos.com/os/docs/latest/booting-on-vagrant.html for further information.
```
puja108: Hmm, we should firstly link to our coreos provisioning stuff and maybe even to onsho before we mention coreos vagrant. (not that I don't want to mention the coreos vagrant option, just want our options to be there, too). But open to other thoughts.

docs/elasticsearch.md
```diff
@@ -0,0 +1,204 @@
...
+- There are three nodes in the cluster, and each node has 2048MB of memory allocated to it.
```
puja108, line 16: do the 2Gb per node have a reason? if yes we should mention that, cause not everyone likes to give away 6 gigs to VMs

docs/elasticsearch.md
```diff
@@ -0,0 +1,204 @@
...
+- The `inago` binary, as well as the `elasticsearch` directory from `example`, have been copied to all the nodes in the cluster.
+
+## Starting the Elasticsearch cluster
+We're going to use Inago's `up` command to bring up multiple group slices of the `elasticsearch` group. `up` is equivalent to `submit`, followed by `start`.
+
+Due to the `Conflicts` statement in the unit file for the group, each group slice will be automatically scheduled onto separate machines.
```
puja108: I'd say `onto separate machines` -> `onto a separate machine` (we're talking about "each group slice")

docs/elasticsearch.md
```diff
@@ -0,0 +1,204 @@
...
+  "initializing_shards" : 0,
+  "unassigned_shards" : 0,
+  "delayed_unassigned_shards" : 0,
+  "number_of_pending_tasks" : 0,
+  "number_of_in_flight_fetch" : 0,
+  "task_max_waiting_in_queue_millis" : 0,
+  "active_shards_percent_as_number" : 100.0
+}
+```
+The cluster status is `green`, and we have three nodes running. Hurrah!
```
puja108, line 90: or `Huzzah`? ;)

docs/elasticsearch.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+
+However, this isn't the latest version of Elasticsearch! We want to upgrade it to 2.3.
+
+We need to modify the line to read:
+```
+Environment="IMAGE=elasticsearch:2.3"
+```
+This will update us to use the later version of the Elasticsearch Docker image.
```
puja108: `update us`? somehow sounds off

puja108:
Made some comments inline.

Also wanted to brush two more issues that I'm not 100% sure about:
1. We say Inago is especially fit for managing groups of unit files. However, what we're showing here is actually only slices of one single unit file, not a group. Would it be a lot of work to extend this to a complete ELK?
2. Why do we need to run elasticsearch on fleet with Inago? What do we gain compared to running it on a higher level orchestration framework like appd or k8s etc.?

docs/elasticsearch.md
```diff
@@ -0,0 +1,204 @@
+# Elasticsearch
```
xh3b4sd, line 1: Is there a reason why this is not in the `example` directory? IMO we should put examples in there. We can put the unit files in there so people can use them directly when checking out the repository. A README for each example explains then how it works. What do you guys think?

docs/elasticsearch.md
```diff
@@ -0,0 +1,204 @@
...
+
+With Inago, it's a piece of cake to deploy an Elasticsearch cluster,
+as well as updating it without any downtime or data loss.
```
xh3b4sd, line 4: The line breaks feel arbitrary. Sometimes they are sufficient, sometimes not. I would love to see some consistent and reasonable line breaks to shorten lines.

docs/elasticsearch.md
```diff
@@ -0,0 +1,204 @@
...
+
+It should be noted that this is a demo, and there are dangers to using it in a production environment without further testing.
+Saying that, if you want to do so, and have questions, the team is always available on IRC (`#giantswarm` on `freenode`) <3
```
xh3b4sd, line 7: <3

docs/elasticsearch.md
xh3b4sd: +1 We should do this for marketing bullshit reasons. At the end it shouldn't matter where the examples are deployed for the showcase anyway.

docs/elasticsearch.md
xh3b4sd, line 1: I just saw at the end of the PR that there is a unit file in the examples directory. It is weird to have things that belong together in different places. in `docs` there is this guide here, and in `example` there is the unit file. 😕 

I ignore the inconsistency of singular vs plural in the directory naming of docs vs example because this has nothing to do with this PR. 😒

docs/elasticsearch.md
```diff
@@ -0,0 +1,204 @@
...
+
+Due to the `Conflicts` statement in the unit file for the group, each group slice will be automatically scheduled onto separate machines.
+```
+core@core-01 ~ $ ./inagoctl up elasticsearch 3
+2016-04-06 16:57:39.644 | INFO     | context.Background: Succeeded to submit group 'elasticsearch'.
+2016-04-06 16:57:46.656 | INFO     | context.Background: Succeeded to start 3 slices for group 'elasticsearch': [1f4 8f1 9e9].
+```
+As you can see, the `elasticsearch` group has been submitted and started. Each group slice has an ID, which can be seen in the output.
+
+We can use the `status` command to inspect the state of the group. The _verbose_ flag is used so that the hash of the unit file is also printed. This will be interesting later, after we upgrade Elasticsearch. You can also see the the group slices have been scheduled onto multiple machines.
```
xh3b4sd: I don't like such long lines. Anyway `the the` -> `the`.

docs/elasticsearch.md
```diff
@@ -0,0 +1,204 @@
...
+```
+core@core-01 ~ $ ./inagoctl status -v elasticsearch
+Group              Units                      FDState   FCState   SAState  Hash                                      IP            Machine
+
+elasticsearch@2cf  elasticsearch@2cf.service  launched  launched  active   07969bf8caa1a7f8c73b35bb826229425d3264f6  172.17.8.103  1564e0d3c4ab4aadbe76b24cea90c62f
+elasticsearch@445  elasticsearch@445.service  launched  launched  active   07969bf8caa1a7f8c73b35bb826229425d3264f6  172.17.8.102  2dbbe125e237410bb946e7b0f6b95e4c
+elasticsearch@891  elasticsearch@891.service  launched  launched  active   07969bf8caa1a7f8c73b35bb826229425d3264f6  172.17.8.101  b89277d82f6a425eb83abb39c00485bc
+```
+
+We can also check Elasticsearch via its own API.
```
xh3b4sd, line 69: Do we want to note that the endpoint in the curl request needs to be modified according to the personal setup?

docs/elasticsearch.md
```diff
@@ -0,0 +1,204 @@
...
+```
+The cluster status is `green`, and we have three nodes running. Hurrah!
+
+We'll add an index, and a document to that index, too.
+```
+core@core-02 ~ $ curl -XPUT '172.17.8.102:9200/inago-example-test-index?pretty'
+{
+  "acknowledged" : true
+}
+core@core-02 ~ $ curl -XPUT 'http://172.17.8.103:9200/inago-example-test-index/external/1?pretty' -d '{"nanana": "batman!"}'
```
xh3b4sd, line 98: There are some `nananas` missing.

docs/elasticsearch.md
```diff
@@ -0,0 +1,204 @@
...
+We need to modify the line to read:
+```
+Environment="IMAGE=elasticsearch:2.3"
+```
+This will update us to use the later version of the Elasticsearch Docker image.
+
+Next, we're going to use the `update` command from Inago to perform the update of the Elasticsearch cluster.
+```
+core@core-01 ~ $ ./inagoctl update elasticsearch --max-growth=1 --min-alive=2 --ready-secs=60
+2016-04-06 17:11:54.026 | INFO     | context.Background: Succeeded to update 3 slices for group 'elasticsearch': [184 372 88f].
```
xh3b4sd, line 163: IMO there should be a little bit more info logging. #notinthispr

docs/elasticsearch.md
```diff
@@ -0,0 +1,204 @@
...
+```
+The arguments used here mean that Inago is allowed to create one additional slice during the update (meaning there will be no more than four instances of Elasticsearch running at any time), and that at least two instances of Elasticsearch have to be running at any time.
+
+The `ready-secs` flag determines how long to wait between rounds of updating. This value gives Elasticsearch enough time to replicate any necessary data.
```
xh3b4sd: We should note that this is highly dependent on the personal setup.

xh3b4sd:
Responding to @puja108's notes.

> We say Inago is especially fit for managing groups of unit files. However, what we're showing here is actually only slices of one single unit file, not a group. 

This shows that the concepts are still not clear. In fact the single unit file represents a group. It is only that this group only consists of one unit. We thought this is sufficient for an example to show what is possible with Inago. 

> Would it be a lot of work to extend this to a complete ELK?

IMO this examples show some functionality. When we want to have some recipes that people simply can run then we can do that. It seems to not be clear what we want to achieve here. Should the focus be on Inago or some other open source tool? This question is important to answer. As I understood it the goal of this task here was to show functionality of Inago. Showing the use of third party open source tools for my understanding is a different thing. Further such recipes are going to be massive because of the amount of components needed to setup things. We cannot simply use our internal images in all cases and need to prepare lots of things to get this working for the public. 

> Why do we need to run elasticsearch on fleet with Inago? What do we gain compared to running it on a higher level orchestration framework like appd or k8s etc.?

For me this is flexibility and easy. `appd` and `k8s` require additional setups and more preparation work. Then the result would be nearly the same I guess. Further you only have unit files. Plain and simple. No other fancy definition needs to be known. So Inago provides the same with less overhead. When it later comes to more sophisticated use cases like user management or locking actions when multiple people work together on something Inago is not going to be sufficient anymore with its current state. And here even `appd` then requires more tools around it to accomplish such things. Personally I cannot say much about `k8s` in this direction.

docs/elasticsearch.md
puja108: +1 (and the amount of data)

puja108:
@xh3b4sd I do get your points and you are right with your response, if just looking at showing Inago functionality.

I think it is still not clear what we want to show with this example. If it's just functionality then this is totally fine.

My thing is that from a product perspective we need to be clear about what Inago will be for and what it won't be for. From my point of view Inago won't be a developer-facing tool, but rather a ops-facing one, something that helps ops deploy infrastructure (modules). That is why I kind of want external people who look at this repo to know why we chose this example.

To be clear, I'm ok with having this example as is, but I'd like to get some more opinions on the general goal that we're after with this. @teemow @zeisss?

xh3b4sd:
> I think it is still not clear what we want to show with this example.

+1

> From my point of view Inago won't be a developer-facing tool

To be honest. When I think about having some application I develop I would consider Inago to deploy my stuff. It is easy and it works quite well. The only pain is hustling with plain unit files. So some configuration management is missing. What I personally would like to prevent when wanting to get something done is certain amount of required knowledge. For appd and k8s I need to know how to setup and how to use and how it fits my stuff. Both tools are living in there own universe where devil hides in the dirty details. I really don't care about such things when I want to deploy my application. The same applies to Inago but the Inago universe is way more straight forward. But again, when it comes to monitoring, logging, user-management, etc. Inago is not a good fit. It is only that I certainly don't care about these things when being a single developer only wanting to orchestrate his application.

docs/elasticsearch.md
teemow: I wouldn't use Mayu/Onsho for that. Vagrant is fine or docker/systemd-nspawn/conair.

puja108: I'd change this then to say sth like `you can use this on any standard fleet cluster, if you want to try it out locally, try the vagrant coreos stuff`

docs/elasticsearch.md
JosephSalisbury: See https://github.com/giantswarm/inago/issues/217

docs/elasticsearch.md
JosephSalisbury, line 163: See https://github.com/giantswarm/inago/issues/218

docs/elasticsearch.md
JosephSalisbury, line 1: I modelled the structure on the Kubernetes example, with groups in `example` and walkthrough in `docs`.
So for now, I'd like to keep it as is.

For discussion on the locations, see: https://github.com/giantswarm/inago/issues/219
For `docs` vs. `example` naming inconsistency, see: https://github.com/giantswarm/inago/issues/220

docs/elasticsearch.md
JosephSalisbury, line 4: I've attempted to make them more uniform. Feel free to point out any other points where you dislike it.

docs/elasticsearch.md
JosephSalisbury, line 69: Done.

JosephSalisbury:
yo @puja108 / @xh3b4sd - mind having another look? i think i've addressed all the points you raised

docs/elasticsearch.md
xh3b4sd, line 4: It is nothing critical but just have a look at line 3 and 6. The differences are across the whole document.

xh3b4sd:
Lets simply go with this. LGTM

PR #233, opened 2016-04-14
adds an int test that updates a unit with only one slice

Nesurion:
we don't have a test for this case yet.

### DIFF
--- a/int-tests/007-update.t
+++ b/int-tests/007-update.t
@@ -22,6 +22,5 @@ Modify unit and perform update
 
 Shut down
   $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} stop $GROUP >050.out 2>&1
-  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP -v >061.out 2>&1
   $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP >065.out 2>&1
   $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} destroy $GROUP >070.out 2>&1
--- a/int-tests/007-update.t
+++ b/int-tests/007-update.t
@@ -1,0 +1,29 @@
+This cram test performs the update command on one slice.
+
+Setup unit files
+
+  $ GROUP=007-update
+  $ mkdir $GROUP
+  $ echo "[Unit]\nDescription=Unit 1\n[Service]\nExecStart=/bin/bash -c 'while true; do echo Hello %n; sleep 10; done'\n" > $GROUP/$GROUP-1-unit@.service
+  $ echo "[Unit]\nDescription=Unit 2\n[Service]\nExecStart=/bin/bash -c 'while true; do echo Hello %n; sleep 10; done'\n[X-Fleet]\nMachineOf=$GROUP-1-unit@%i.service\n" > $GROUP/$GROUP-2-unit@.service
+
+Submit units
+
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} submit $GROUP 1 >010.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP -v >021.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP >025.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} start $GROUP >030.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP -v >041.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP > 045.out  2>&1
+  $ sleep 10
+
+Modify unit and perform update
+  $ echo "[Unit]\nDescription=Unit 1 (CHANGED)\n[Service]\nExecStart=/bin/bash -c 'while true; do echo Hello %n; sleep 10; done'\n" > $GROUP/$GROUP-1-unit@.service
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} update $GROUP --max-growth=1 --min-alive=1 > 48.out 2>&1
+  $ sleep 10
+
+Shut down
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} stop $GROUP >050.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP -v >061.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP >065.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} destroy $GROUP >070.out 2>&1
--- a/int-tests/007-update.t
+++ b/int-tests/007-update.t
@@ -1,4 +1,6 @@
-This cram test performs the update command on one slice.
+This cram test performs the update command on one slice. We run this test
+to make sure that handling the update of just one slice works aswell.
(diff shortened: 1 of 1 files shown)


Nesurion:
@giantswarm/cena RFR

int-tests/007-update.t
```diff
@@ -0,0 +1,29 @@
+This cram test performs the update command on one slice.
```
xh3b4sd: Can we add more information why we care about this special case?

xh3b4sd: Further it looks like this test is about a group having two units and one slice of the group.

int-tests/007-update.t
```diff
@@ -0,0 +1,29 @@
...
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP >025.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} start $GROUP >030.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP -v >041.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP > 045.out  2>&1
+  $ sleep 10
+
+Modify unit and perform update
+  $ echo "[Unit]\nDescription=Unit 1 (CHANGED)\n[Service]\nExecStart=/bin/bash -c 'while true; do echo Hello %n; sleep 10; done'\n" > $GROUP/$GROUP-1-unit@.service
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} update $GROUP --max-growth=1 --min-alive=1 > 48.out 2>&1
+  $ sleep 10
```
xh3b4sd, line 23: Why is there a sleep in there?

int-tests/007-update.t
```diff
@@ -0,0 +1,29 @@
...
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP -v >021.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP >025.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} start $GROUP >030.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP -v >041.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP > 045.out  2>&1
+  $ sleep 10
+
+Modify unit and perform update
+  $ echo "[Unit]\nDescription=Unit 1 (CHANGED)\n[Service]\nExecStart=/bin/bash -c 'while true; do echo Hello %n; sleep 10; done'\n" > $GROUP/$GROUP-1-unit@.service
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} update $GROUP --max-growth=1 --min-alive=1 > 48.out 2>&1
```
xh3b4sd: Can we maybe decrease the ready secs? This can also be the cause of int tests being so damn slow. We probably should turn that down to 1 or 2 everywhere. Depending on the number of int tests this could boost the whole thing a bit.

int-tests/007-update.t
```diff
@@ -0,0 +1,29 @@
...
+  $ echo "[Unit]\nDescription=Unit 2\n[Service]\nExecStart=/bin/bash -c 'while true; do echo Hello %n; sleep 10; done'\n[X-Fleet]\nMachineOf=$GROUP-1-unit@%i.service\n" > $GROUP/$GROUP-2-unit@.service
+
+Submit units
+
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} submit $GROUP 1 >010.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP -v >021.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP >025.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} start $GROUP >030.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP -v >041.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP > 045.out  2>&1
```
xh3b4sd, line 17: Why are we doing status calls always twice?

int-tests/007-update.t
Nesurion, line 23: to be honest: no idea. copy pasted that from https://github.com/giantswarm/inago/blob/master/int-tests/004-update.t

int-tests/007-update.t
Nesurion, line 17: I guess to validate that `-v` works aswell. But lets change that so its more like a golden-path test

xh3b4sd, line 17: +1 to have a separate test for that.

int-tests/007-update.t
xh3b4sd, line 23: @JosephSalisbury @zeisss maybe have more information? I would vote to remove it. Only makes the tests slow.

int-tests/007-update.t
Nesurion: number of units don't matter here, number of slices do.

xh3b4sd: Anyway it clarifies what the setup looks like. There is enormous potential to get the concepts wrong and create misunderstanding IMO.

Nesurion:
@xh3b4sd could you take another look?

int-tests/007-update.t
```diff
@@ -0,0 +1,27 @@
...
+
+Modify unit and perform update
+  $ echo "[Unit]\nDescription=Unit 1 (CHANGED)\n[Service]\nExecStart=/bin/bash -c 'while true; do echo Hello %n; sleep 10; done'\n" > $GROUP/$GROUP-1-unit@.service
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} update $GROUP --max-growth=1 --min-alive=1 --ready-secs=2 > 48.out 2>&1
+
+Shut down
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} stop $GROUP >050.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP -v >061.out 2>&1
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} status $GROUP >065.out 2>&1
```
xh3b4sd, line 27: Duplicated status call.

xh3b4sd:
I like that. Would you mind tuning the other tests as well?

int-tests/007-update.t
Nesurion, line 27: fixing it now.
_don't do things like this when your head is already at lunch_

Nesurion:
Can do, but I'd like to get @zeisss and @JosephSalisbury opinion first. Maybe they remember why we put in the sleep in the first place.

JosephSalisbury:
If i remember correctly, we had the sleep statements to give the system a beat to get consistent. If you get consistent good runs without them, we don't need them.

Nesurion:
oh yeah I remember. Running the int-tests without the sleep resultet in random failing tests, which was super annoying.
I'll add them back in.

xh3b4sd:
But wait. Why is that then? Shouldn't we understand and find out the root cause?

JosephSalisbury:
I think it might boil down to 'lol fleet', but looking into it further would be a good bug hunting expedition.

Nesurion:
@giantswarm/cena RFR. I'll test things with lower or no sleep times in separate PR.

xh3b4sd:
Was this actually related to a known bug we fixed? I would maybe point that out in the cram description. Besides that, LGTM.

Nesurion:
I thought it was a bug, when it was just my local fleet being broken. Nevertheless I wanted to have a test that tests the update command against a single slice.

PR #235, opened 2016-04-14
info logging for updates

### LINKED ISSUE #218: Increase info logging output during update
It would be good to see when an instance is created / destroyed

xh3b4sd:
fixes #218 

This is how it looks like.

```
core@core-01 ~ $ ./inagoctl update test-group
2016-04-14 11:04:10.747 | INFO     | context.Background.WithValue("task-id", "8596262e-43f4-4175-bb1f-86b907e9c5e1").WithValue("slice ID", "877"): controller: adding units
2016-04-14 11:04:55.779 | INFO     | context.Background.WithValue("task-id", "8596262e-43f4-4175-bb1f-86b907e9c5e1").WithValue("slice ID", "877"): controller: removing units
2016-04-14 11:05:04.810 | INFO     | context.Background: Succeeded to update 1 slice for group 'test-group': [877].
```

RFR @JosephSalisbury @zeisss @Nesurion


Nesurion:
Do we want the task ID in there? My guess would be that it's not relevant on an info log level.

The int tests need some love, after changing the output of a command (update in this case).

xh3b4sd:
I am also not a fan of the context information in there. But what to do? Would be cool to have it consistently.

xh3b4sd:
Fishing for LGTM

controller/update.go
```diff
@@ -144,6 +144,8 @@ func (c controller) addFirst(ctx context.Context, req Request, opts UpdateOption
 }
 
 func (c controller) runAddWorker(ctx context.Context, req Request, opts UpdateOptions) (Request, error) {
+	c.Config.Logger.Info(nil, "controller: adding units %v", req.SliceIDs)
```
JosephSalisbury: Why aren't you using ctx here?

xh3b4sd: Please consider reading https://github.com/giantswarm/inago/pull/235#issuecomment-209897009

I thought it is not necessary for info logging. I also think it is ugly as fuck. I am totally open on that. What would you like to see?

JosephSalisbury: Bah, I shouldn't respond to comments when I'm in the pub. It's very useful to have the context information in. If the issue is that logging is ugly, we can address that separately.

xh3b4sd: @Nesurion argumented the additional contextual information is too much for the info output. I agree with that. So I removed the context. Now there are all information that are needed. In case you want to have more information you can use the debug log level, that is `-v`. I don't think the context is useful here. I even think the context is very much useless because of its limited design, but maybe I simply get something wrong. Anyway, I am open about this. Tell me how it should be.

JosephSalisbury: Ah, I see. Yeah, I can get behind that. Using nil in this scenario makes sense.

Nesurion:
I have to change my mind on this. The context not being useful was mostly due to the fact that we didn't pass the sliceIDs into goroutines, but rather accessed a shared var containing the sliceID => all actions that are done inside the go routine use the same sliceID.

in https://github.com/giantswarm/inago/blob/max-growth-bug-fix/controller/update.go#L340 we fixed that. So the debug output makes sense again.

JosephSalisbury:
Then, let's use the proper context! :D

xh3b4sd:
> I have to change my mind on this.

I do not.

> The context not being useful was mostly due to the fact that we didn't pass the sliceIDs into goroutines, but rather accessed a shared var containing the sliceID => all actions that are done inside the go routine use the same sliceID.

@Nesurion you pointed to https://github.com/giantswarm/inago/blob/max-growth-bug-fix/controller/update.go#L340. I guess the thing is already merged. It is wrong though. Sorry I did not realize earlier. The slice ID is now in the goroutine, but the request is forwarded, not the slice ID. So Instead of passing the slice ID properly to the goroutines, this should happen with the request being processed instead, because this one is forwarded to the deeper methods. 

Anyway I will add back the context even it makes no sense to me. #democracy

Nesurion:
@xh3b4sd not merged yet, I want to figure out the bad codecov test score first.
regarding the ctx. You're right, we tested things with a slightly different code, that is: we created a new ctx inside the goroutine with the passed sliceID.

Nesurion:
LGTM

Issue #220, opened 2016-04-08
`docs` and `example` are not named consistently


JosephSalisbury:
It should be `docs` and `examples` or `doc` and `example`.

I'd say the former (plurals), personally.

zeisss:
+1 Plural. Thats what we used in our https://github.com/giantswarm/example-opensource-repo anyways.

puja108:
done

PR #239, opened 2016-04-14
updating k8s example for use in blog post

puja108:
first commit:
based on releaseit output
v1.0.1 currently
still about to move images to public dockerhub


puja108:
Works well with 1.0.1, 1.0.2, 1.2.0 and most probably even 1.2.2 (I didn't upload any other k8s version to dockerhub). Goal is to show 1.2.0 and then a rolling update of the nodes to 1.2.2. For now without any scheduled pods, but might add a tiny pod in there so you can see that it survives (might need to tweak wait time a bit).

puja108:
More todos:
- [ ] unit files might need some cleaning up
- [x] not sure if `setup-network-env` needs to be in the master group or should be by itself (@teemow ?)

puja108:
This is now working incl rolling updates with a running (scaled) pod and in line with the soon to be published inago blog post. Should be clean enough for a first publication.

RFR @giantswarm/cena

docs/k8s.md
```diff
@@ -1,47 +1,151 @@
...
+
+Now, we can schedule the nodes. As we have a 3 node cluster we'll be running 3 slices of nodes:
+
+```
+$ inagoctl up k8s-node 3
+2016-04-14 15:41:26.824 | INFO     | context.Background: Succeeded to submit group 'k8s-node'.
+2016-04-14 15:41:36.847 | INFO     | context.Background: Succeeded to start 3 slices for group 'k8s-node'
+```
+
+And we're done!
```
xh3b4sd, line 45: I am wondering why we need three different groups. I want to execute one command and magic should happen. What are we missing to do this? Why do we need three commands? One would be super awesome.

docs/k8s.md
```diff
@@ -1,47 +1,151 @@
...
+
+Looking closely at the update command above we can inspect some of its options. `--max-growth=0` tells Inago to not start any additional instances, which means it has to remove an instance to be able to start a new one. In this example this is due to the fact that we don't have any free machine left. `--min-alive=2` tells Inago to keep at least two instances of the group alive during the update process. the `ready-secs` flag determines how long Inago waits between rounds of updating. This gives the new node a bit of time to start replicating that pod we scheduled.
+
+Watching `kubectl get deployments` during the update process, we would see that available instances of our hello-node pod shortly go down to 2, while a Kubernetes node is being updated and then quickly come up again to 3.
+
+### Updating the Master
+
+Now sadly, we didn't use an HA deployment of the Kubernetes master, so Inago's update command isn't of help much here. Accordingly, we will have a short downtime of our API while we update this group. Luckily, Kubernetes nodes and their pods don't rely on the API server that much and will usually stay running while we update the master.
```
xh3b4sd: `isn't of help` is weird.

docs/k8s.md
```diff
@@ -1,47 +1,151 @@
...
+
+$ inagoctl destroy k8s-master
+2016-04-14 16:32:27.384 | INFO     | context.Background: Succeeded to destroy group 'k8s-master'.
+
+$ inagoctl up k8s-master
+2016-04-14 16:35:23.503 | INFO     | context.Background: Succeeded to submit group 'k8s-master'.
+2016-04-14 16:36:30.600 | INFO     | context.Background: Succeeded to start group 'k8s-master'.
 ```
+
+A quick look at `kubectl version` will tell us that we're now running a 1.2.2 server.
```
xh3b4sd: Based on a gut feeling I miss some "ending words". Here the post is suddenly over and I have the feeling something is missing. Nobody said it is over. Nobody said goodby. Does this make sense?

xh3b4sd:
Minor comments. The post is cool. I only still don't like the different locations of docs and unit files.

docs/k8s.md
puja108, line 45: The network one can be integrated with the master, but I thought it's cleaner outside as it's one-off and actually does not belong to k8s itself.

master and nodes need to be deployed separately as the master in this example is not HA (and not sliced).

Also, you would want some control over master separately from nodes as they are two different things. Even with an HA mster you would maybe want only 3 slices of the master, but you might want 10 nodes.

docs/k8s.md
puja108: true, changed to `currently won't work`

docs/k8s.md
puja108: yes, I agree, actually have some in the blog post, but kinda forgot them here. will add.

docs/k8s.md
xh3b4sd, line 45: I see the group differentiation makes sense. Anyway I think even for multiple groups one `up` should be enough. Having that said there would be a need for group specific configuration. Then it starts to feel not that handy anymore.

docs/k8s.md
```diff
@@ -1,47 +1,151 @@
 # Kubernetes
 
 With Inago you can easily deploy Kubernetes on a CoreOS cluster.
-The `example` folder contains a `k8s` example. Kubernetes is split into
+The [example folder](https://github.com/giantswarm/inago/tree/master/example) contains a simple [k8s example](https://github.com/giantswarm/inago/tree/master/example/k8s). The Kubernetes deployment is split into
 two groups: `k8s-master` and `k8s-node`.
 
 ## Note
 The deployment sets up a Kubernetes cluster, which uses insecure communication
```
zeisss: I would change this `which`.

> The deployment sets up a Kubernetes cluster with insecure communication between the components.

docs/k8s.md
```diff
@@ -1,47 +1,151 @@
...
 between the components. Use this only as a Demo or POC. If you want to deploy a
 Kubernetes cluster for production please modify the example, so secure
-communication is used. [Cluster TLS using OpenSSL](https://coreos.com/Kubernetes/docs/latest/openssl.html)
+communication is used: [Cluster TLS using OpenSSL](https://coreos.com/Kubernetes/docs/latest/openssl.html).
+Further, this deployment uses a single Kubernetes master, so the API is not highly available. For HA deployments of Kubernetes check out: [Building High-Availability Clusters](http://kubernetes.io/docs/admin/high-availability/).
```
zeisss: I wouldn't put a double colon here.

docs/k8s.md
puja108: done

docs/k8s.md
puja108: ack

puja108:
@xh3b4sd is my answer to your 1 vs many groups comment acceptable, if LGTY

puja108:
fixes https://github.com/giantswarm/inago/issues/220

xh3b4sd:
Yeah, discussed. LGTM

PR #240, opened 2016-04-15
trim trailing slash from group

### LINKED ISSUE #210: If a group name is suffixed with '/', remove the slash
I use tab completion a lot, it would be nice if I didn't have to backspace

xh3b4sd:
fixes #210 

See tests. 

RFR @JosephSalisbury @Nesurion @zeisss

### DIFF
--- a/controller/request.go
+++ b/controller/request.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"fmt"
 	"regexp"
+	"strings"
 
 	"golang.org/x/net/context"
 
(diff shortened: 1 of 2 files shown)


Nesurion:
LGTM

JosephSalisbury:
My original idea was just to do this in the `cli` package - why do we want it in the controller?

xh3b4sd:
Makes no sense to me in the CLI. Would be needed in every damn place. Anyway, the "fix" is only a PR away.

JosephSalisbury:
Works for me for now either way.

PR #242, opened 2016-04-15
fix max-growth bug

### LINKED ISSUE #213: Update can cause an extraneous slice to be created
@xh3b4sd and I had a case today where we had two slices running, and an update with min-alive=1, max-growth=0 presented an update like so:
1. [339, 591]
2. [591]
3. [0a0, 591, 72e]
4. [0a0, 72e]

It would seem that at 3, 2 additional slices were made. With max-growth=0, we would expect only one slice to have been created at this point.

Nesurion:
See #213

We discovered that the calculation if we are allowed to add or remove slices was faulty. The PR fixes #213 by changing the calculation formula.

### DIFF
--- a/controller/update.go
+++ b/controller/update.go
@@ -72,20 +72,20 @@ func (c controller) getNumRunningSlices(ctx context.Context, req Request) (int, 
 	return len(sliceIDs), nil
 }
 
-func (c controller) isGroupRemovalAllowed(ctx context.Context, req Request, minAlive int, removalInProgress *int64) (bool, error) {
+func (c controller) isGroupRemovalAllowed(ctx context.Context, req Request, minAlive int, removeInProgress *int64) (bool, error) {
 	c.Config.Logger.Debug(ctx, "controller: checking group removal allowed, req: %v", req)
 
 	c.Config.Logger.Debug(
-		ctx, "controller: removalInProgress: %v, minAlive: %v",
-		*removalInProgress, minAlive,
+		ctx, "controller: removeInProgress: %v, minAlive: %v",
+		*removeInProgress, minAlive,
 	)
 
-	if (minAlive - int(*removalInProgress)) > 0 {
-		c.Config.Logger.Debug(ctx, "controller: group removal allowed ((minAlive - int(removalInProgress)) > 0)")
+	if (minAlive - int(*removeInProgress)) > 0 {
+		c.Config.Logger.Debug(ctx, "controller: group removal allowed ((minAlive - int(removeInProgress)) > 0)")
 		return true, nil
 	}
 
-	c.Config.Logger.Debug(ctx, "controller: group removal not allowed ((minAlive - int(removalInProgress)) <= 0)")
+	c.Config.Logger.Debug(ctx, "controller: group removal not allowed ((minAlive - int(removeInProgress)) <= 0)")
 	return false, nil
 }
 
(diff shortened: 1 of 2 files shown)


Nesurion:
@giantswarm/cena RFR. WFM would be super nice

controller/update.go
```diff
@@ -72,47 +72,37 @@ func (c controller) getNumRunningSlices(ctx context.Context, req Request) (int,
...
-	if numRunning > minAlive {
-		c.Config.Logger.Debug(ctx, "controller: group removal allowed (numRunning > minAlive)")
+	if (minAlive - int(*removalInProgress)) > 0 {
+		c.Config.Logger.Debug(ctx, "controller: group removal allowed ((minAlive - int(removalInProgress)) > 0)")
 		return true, nil
 	}
 
-	c.Config.Logger.Debug(ctx, "controller: group removal not allowed (numRunning <= minAlive)")
+	c.Config.Logger.Debug(ctx, "controller: group removal not allowed ((minAlive - int(removalInProgress)) <= 0)")
 	return false, nil
```
xh3b4sd, line 89: Do you thing we still need this methods?

controller/update.go
```diff
@@ -348,30 +337,34 @@ func (c controller) UpdateWithStrategy(ctx context.Context, req Request, opts Up
...
 
 				break
 			}
 
 			c.Config.Logger.Debug(ctx, "controller: attempting to remove slice: %v", sliceID)
 			// remove
-			minAlive := opts.MinAlive + int(removeInProgress)
+			// we are only allowed to remove if the number of minAlive slices
+			// minus the ones, that are beeing removed right now is greater than 0
+			//=> (minAlive - int(removalInProgress)) > 0)
```
xh3b4sd: `removalInProgress` -> `removeInProgress`

controller/update.go
Nesurion, line 89: I'd like to keep it in, as it was useful while writing the fix. I guess it could come in handy at some point in the future

xh3b4sd:
LGTM

controller/update.go
```diff
@@ -348,30 +337,34 @@ func (c controller) UpdateWithStrategy(ctx context.Context, req Request, opts Up
 
 					atomic.AddInt64(&addInProgress, -1)
 					done <- struct{}{}
-				}()
+				}(ctx)
```
xh3b4sd, line 340: Looks like I was not clear enough. `newReq` is the one we need to pass into the goroutine scope as well.

controller/update.go
```diff
@@ -312,29 +301,29 @@ func (c controller) UpdateWithStrategy(ctx context.Context, req Request, opts Up
...
 				return maskAny(err)
 			}
 			if ok {
-				go func() {
+				// we increase the addInProgress counter before starting the goroutine
+				// to avoid a race condition in the allowed calculation
+				atomic.AddInt64(&addInProgress, 1)
+				go func(ctx context.Context) {
 					ctx = context.WithValue(ctx, "add slice", sliceID)
 					c.Config.Logger.Debug(ctx, "controller: starting to add slice: %v", sliceID)
```
xh3b4sd, line 326: Here `newReq.SliceIDs` should be used for logging.

controller/update.go
```diff
@@ -348,30 +337,34 @@ func (c controller) UpdateWithStrategy(ctx context.Context, req Request, opts Up
...
+			ok, err = c.isGroupRemovalAllowed(ctx, currentSliceReq, opts.MinAlive, &removeInProgress)
 			if err != nil {
 				return maskAny(err)
 			}
 			if ok {
-				go func() {
+				// we increase the removeInProgress counter before starting the goroutine
+				// to avoid a race condition in the allowed calculation
+				atomic.AddInt64(&removeInProgress, 1)
+				go func(ctx context.Context) {
```
xh3b4sd, line 365: Same comments as above apply also to this goroutine.

Nesurion, line 365: @xh3b4sd please see https://github.com/giantswarm/inago/pull/243

PR #243, opened 2016-04-18
show correct sliceID in logs when running goroutines

Nesurion:
Currently the request is not passed into the goroutine and thus the log outputs inside the goroutines only show the latest sliceID, not the one they are executed on.
This PR passes the request into the goroutines, enabling them to log the correct sliceID.

### DIFF
--- a/controller/update.go
+++ b/controller/update.go
@@ -325,14 +325,14 @@ func (c controller) UpdateWithStrategy(ctx context.Context, req Request, opts Up
 					ctx := context.WithValue(ctx, "add slice", req.SliceIDs)
 					c.Config.Logger.Debug(ctx, "controller: starting to add slice: %v", req.SliceIDs)
 
-					newSliceIDs, err := c.addFirst(ctx, newReq, opts)
+					newSliceIDs, err := c.addFirst(ctx, req, opts)
 					if err != nil {
 						fail <- maskAny(err)
 						return
 					}
 
 					currentSliceIDsMutex.Lock()
-					currentSliceIDs = c.updateCurrentSliceIDs(ctx, currentSliceIDs, newReq.SliceIDs, newSliceIDs)
+					currentSliceIDs = c.updateCurrentSliceIDs(ctx, currentSliceIDs, req.SliceIDs, newSliceIDs)
 					currentSliceIDsMutex.Unlock()
 
 					atomic.AddInt64(&addInProgress, -1)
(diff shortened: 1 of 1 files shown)


Nesurion:
Debug log lines look like this:
`2016-04-18 10:04:55.734 | DEBUG    | context.Background.WithValue("task-id", "8e3fbffc-1ee7-49fd-9854-778e90f30e1d").WithValue("add slice", []string{"63c"}): controller: running addFirst`

I dislike the `[]string{"63c"}`part. Just showing the ID would be nicer here. Not sure how to do it though :/

controller/update.go
```diff
@@ -321,9 +321,9 @@ func (c controller) UpdateWithStrategy(ctx context.Context, req Request, opts Up
...
 				// to avoid a race condition in the allowed calculation
 				atomic.AddInt64(&addInProgress, 1)
-				go func(ctx context.Context) {
-					ctx = context.WithValue(ctx, "add slice", sliceID)
-					c.Config.Logger.Debug(ctx, "controller: starting to add slice: %v", sliceID)
+				go func(req Request) {
+					ctx := context.WithValue(ctx, "add slice", req.SliceIDs)
+					c.Config.Logger.Debug(ctx, "controller: starting to add slice: %v", req.SliceIDs)
 
 					newSliceIDs, err := c.addFirst(ctx, newReq, opts)
```
xh3b4sd: The whole point of this fix is actually passing the request properly to the goroutine AND using it. So `newReq` -> `req`. The issues was not the logging, but the potential raise conditions during updates.

controller/update.go
```diff
@@ -362,9 +362,9 @@ func (c controller) UpdateWithStrategy(ctx context.Context, req Request, opts Up
...
 				// to avoid a race condition in the allowed calculation
 				atomic.AddInt64(&removeInProgress, 1)
-				go func(ctx context.Context) {
-					ctx = context.WithValue(ctx, "remove slice", sliceID)
-					c.Config.Logger.Debug(ctx, "controller: starting to remove slice: %v", sliceID)
+				go func(req Request) {
+					ctx := context.WithValue(ctx, "remove slice", req.SliceIDs)
+					c.Config.Logger.Debug(ctx, "controller: starting to remove slice: %v", req.SliceIDs)
 
 					newSliceIDs, err := c.removeFirst(ctx, newReq, opts)
```
xh3b4sd: Same here.

xh3b4sd:
I dislike the visualization of the context as well. I like the log metadata from appd way more. There is simply a map that we extend as we like. Here the context is only able to carry around one single key value pair. This could be indeed a map itself, but meh.

JosephSalisbury:
Contexts can carry multiple key-values, you need to use [WithValue](https://godoc.org/golang.org/x/net/context#WithValue) multiple times.

xh3b4sd:
I see. https://github.com/golang/net/blob/master/context/pre_go17.go#L295

JosephSalisbury:
Ack, https://github.com/golang/net/blob/e45385e9b226f570b1f086bf287b25d3d4117776/context/pre_go17.go#L286 each context has a parent, and has one key value pair. I get it.

Nesurion:
another RFR @giantswarm/cena

xh3b4sd:
LGTM

PR #244, opened 2016-04-18
file system tests

### LINKED ISSUE #135: Add unit tests to filesystem package

xh3b4sd:
fixes #135 

This tests the memory implementation of the file system package. I am not sure we want to test the OS implementation against a real OS. Opinions? 

RFC @JosephSalisbury @Nesurion @zeisss

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,7 @@ test: $(SOURCE) VERSION .gobuild
 	@echo Testing inside Docker container for $(GOOS)/$(GOARCH)
 	docker run \
 	    --rm \
+	    -t \
 	    -v $(shell pwd):/usr/code \
 	    -e GOPATH=/usr/code/.gobuild \
 	    -e GOOS=$(GOOS) \
@@ -88,7 +89,7 @@ ci-build: $(SOURCE) VERSION .gobuild
 ci-test: $(SOURCE) VERSION .gobuild
 	echo Testing for $(GOOS)/$(GOARCH)
 	$(TEST_COMMAND)
-	
+
 # Use with `GOOS=linux FLEET_ENDPOINT=http://192.168.99.1:49153/ make int-test`
 # Set fleet endpoint to a fleet API endpoint available to the container.
 
@@ -102,7 +103,7 @@ int-test: $(BIN) $(INT_TESTS)
 	@echo Starting CoreOS integration test machine
 	cd $(VAGRANT_PATH) && vagrant up
 	sleep 10
-	
+
 	@echo Starting ssh-agent container
 	-docker run \
 	-d \
@@ -114,9 +115,9 @@ int-test: $(BIN) $(INT_TESTS)
 	-v ~/.vagrant.d:/ssh \
 	whilp/ssh-agent:latest \
 	ssh-add /ssh/insecure_private_key
-	
+
 	-FLEET_ENDPOINT=$(FLEET_ENDPOINT) make internal-int-test
-	
+
 	@echo Destroying the ssh-agent container
 	docker rm -f ssh-agent
 	@echo Destroying the integration test machine
@@ -139,7 +140,7 @@ internal-int-test: $(BIN) $(INT_TESTS)
 bin-dist: $(SOURCE) VERSION .gobuild
 	# Remove any old bin-dist or build directories
 	rm -rf bin-dist build
-	
+
 	# Build for all supported OSs
 	for OS in darwin linux; do \
 		rm -f $(BIN); \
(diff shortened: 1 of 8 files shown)


JosephSalisbury:
I have used temp directories and the like for testing this kind of stuff before, I found it worked pretty well. You might want to try it out.

xh3b4sd:
I really don't like tests which dependencies I cannot control.

JosephSalisbury:
I get that :D

Makefile
```diff
@@ -69,6 +69,7 @@ test: $(SOURCE) VERSION .gobuild
 	@echo Testing inside Docker container for $(GOOS)/$(GOARCH)
 	docker run \
 	    --rm \
+	    -t \
```
xh3b4sd, line 72: @Nesurion IIRC you had issues with `-ti`. I added `-t` to be able to abort tests running in containers with ctrl+c. Do you remember if there is an issue with this?

controller/controller_test.go
```diff
@@ -472,8 +472,8 @@ func givenControllerWithConfig(fmc fleetMockConfig) (Controller, *fleetMock) {
 	newControllerConfig.TaskService = newTaskService
 	newControllerConfig.Logger = newLogger
 	newControllerConfig.WaitCount = 1
-	newControllerConfig.WaitSleep = 10 * time.Millisecond
-	newControllerConfig.WaitTimeout = 30 * time.Millisecond
+	newControllerConfig.WaitSleep = 5 * time.Millisecond
+	newControllerConfig.WaitTimeout = 100 * time.Millisecond
```
xh3b4sd, line 476: I had timing issues when running the unit tests locally. This helped.

cli/request.go
```diff
@@ -23,6 +23,9 @@ func readUnitFiles(fs filesystemspec.FileSystem, dir string) (map[string]string,
 		if fileInfo.IsDir() {
 			continue
 		}
+		if !strings.HasPrefix(fileInfo.Name(), dir) {
+			continue
+		}
```
xh3b4sd, line 28: We only want to load unit files. Inago expect unit files to be prefixed, thus I added the check. Had an issue with that. Now it is fixed.

controller/request_test.go
```diff
@@ -39,8 +38,6 @@ func Test_Request_NewRequest(t *testing.T) {
 
 	for i, testCase := range testCases {
 		newRequest := NewRequest(testCase.Input)
-		fmt.Printf("%#v\n", newRequest)
-		fmt.Printf("%#v\n", testCase.Expected)
```
xh3b4sd, line 43: Just removed debug logging. Somebody forgot to remove before merging. Probably even myself.

xh3b4sd:
Good we checked the file system implementation. Was kind of buggy. RFR @JosephSalisbury @Nesurion @zeisss

cli/request.go
zeisss, line 28: so this means we can add a `README.md` next to the unitfiles, right? Nice :)

file-system/fake/error.go
```diff
@@ -10,9 +12,44 @@ var (
...
-// noSuchFileOrDirectoryError. This error is returned in case there cannot
-// any file be found as requested.
-func IsNoSuchFileOrDirectoryError(err error) bool {
+// IsNoSuchFileOrDirectory checks for the given error to be
+// noSuchFileOrDirectoryError. This error is returned in case there cannot any
+// file be found as requested.
+func IsNoSuchFileOrDirectory(err error) bool {
+	cause := errgo.Cause(err)
+
+	if cause == nil {
```
zeisss: are you sure this is needed? I would expect the remaining to work even if `cause` is nil.

zeisss:
LGTM

@JosephSalisbury If you have many tests, doing real file ops is slow. An inmemory implementation helps a lot here and makes it independent of "disk is full" errors etc.

cli/request.go
xh3b4sd, line 28: Yes.

file-system/fake/error.go
xh3b4sd: You are right. Simplified.

Makefile
Nesurion, line 72: Can't remember what the issue with `-ti` was :/
I'll test things now.

Nesurion, line 72: All tests green. Using `-t` LGTM

JosephSalisbury, line 72: hivemind: i removed `-ti` as it caused builder to fail during release (because it can't open a tty for reasons) - it should be fine having this on the test target

xh3b4sd, line 72: Cool, thanks.

PR #245, opened 2016-04-18
Allow removing units if minAlive=0

Nesurion:
In case the min-alive value is set to 0, we allow removing units.

### DIFF
--- a/controller/update.go
+++ b/controller/update.go
@@ -82,7 +82,7 @@ func (c controller) isGroupRemovalAllowed(ctx context.Context, req Request, minA
 	// if minAlive = 0 we don't need to calculate if we can remove slices,
 	// as the user provided us with the information, that killing slices is fine
 	if (minAlive-int(*removeInProgress)) > 0 || minAlive == 0 {
-		c.Config.Logger.Debug(ctx, "controller: group removal allowed ((minAlive - int(removeInProgress)) > 0)")
+		c.Config.Logger.Debug(ctx, "controller: group removal allowed ((minAlive - int(removeInProgress)) > 0) || minAlive == 0")
 		return true, nil
 	}
 
--- a/controller/update.go
+++ b/controller/update.go
@@ -79,8 +79,9 @@ func (c controller) isGroupRemovalAllowed(ctx context.Context, req Request, minA
 		ctx, "controller: removeInProgress: %v, minAlive: %v",
 		*removeInProgress, minAlive,
 	)
-
-	if (minAlive - int(*removeInProgress)) > 0 {
+	// if minAlive = 0 we don't need to calculate if we can remove slices,
+	// as the user provided us with the information, that killing slices is fine
+	if (minAlive-int(*removeInProgress)) > 0 || minAlive == 0 {
 		c.Config.Logger.Debug(ctx, "controller: group removal allowed ((minAlive - int(removeInProgress)) > 0)")
 		return true, nil
 	}
--- a/int-tests/007-min-alive-zero-update.t
+++ b/int-tests/007-min-alive-zero-update.t
@@ -18,7 +18,7 @@ Modify unit
 
 Update unit
 
-  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} update $GROUP --max-growth=0 --min-alive=0 --ready-secs=5
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} update $GROUP --max-growth=0 --min-alive=0 --ready-secs=2
   .*controller: removing units (re)
   .*controller: removing units (re)
   .*controller: adding units (re)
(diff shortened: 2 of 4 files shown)


Nesurion:
RFR @giantswarm/cena

controller/update.go
```diff
@@ -79,8 +79,9 @@ func (c controller) isGroupRemovalAllowed(ctx context.Context, req Request, minA
 		ctx, "controller: removeInProgress: %v, minAlive: %v",
 		*removeInProgress, minAlive,
 	)
-
-	if (minAlive - int(*removeInProgress)) > 0 {
+	// if minAlive = 0 we don't need to calculate if we can remove slices,
+	// as the user provided us with the information, that killing slices is fine
```
xh3b4sd, line 83: `that killing ALL slices`

controller/update.go
```diff
@@ -79,8 +79,9 @@ func (c controller) isGroupRemovalAllowed(ctx context.Context, req Request, minA
...
+	if (minAlive-int(*removeInProgress)) > 0 || minAlive == 0 {
 		c.Config.Logger.Debug(ctx, "controller: group removal allowed ((minAlive - int(removeInProgress)) > 0)")
```
xh3b4sd: We may want to adapt the logging accordingly.

int-tests/007-min-alive-zero-update.t
```diff
@@ -0,0 +1,28 @@
...
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} start $GROUP > /dev/null 2>&1
+  $ sleep 3
+
+Modify unit
+
+  $ echo "[Unit]\nDescription=Unit 1 (CHANGED)\n[Service]\nExecStart=/bin/bash -c 'while true; do echo Hello %n; sleep 10; done'\n" > $GROUP/$GROUP-unit@.service
+
+Update unit
+
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} update $GROUP --max-growth=0 --min-alive=0
```
xh3b4sd: IIRC we discussed turning the ready secs down so the int tests are running faster. Why don't we do this?

xh3b4sd:
Why are codecov results going crazy?

Nesurion:
But in the `ci.sh` script prevents travis from stopping the build if the unit tests fail.
See https://github.com/giantswarm/inago/pull/246

int-tests/007-min-alive-zero-update.t
Nesurion: we discussed here: https://github.com/giantswarm/inago/pull/233
We didn't investigate further yet why the tests tend to fail without the sleep in there. Will be done in a seperate PR for all int tests.

xh3b4sd: As far as I understood this was about the `sleep`. I am talking about `--ready-secs`. It should be possible to set this down to 5 seconds.

Nesurion:
RFR @giantswarm/cena

int-tests/007-min-alive-zero-update.t
```diff
@@ -0,0 +1,32 @@
+This test checks the behaviourof an update with max-growth zero
```
JosephSalisbury: behaviour of

JosephSalisbury:
yep, LGTM!

int-tests/007-min-alive-zero-update.t
```diff
@@ -0,0 +1,32 @@
...
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} start $GROUP > /dev/null 2>&1
+  $ sleep 3
+
+Modify unit
+
+  $ echo "[Unit]\nDescription=Unit 1 (CHANGED)\n[Service]\nExecStart=/bin/bash -c 'while true; do echo Hello %n; sleep 10; done'\n" > $GROUP/$GROUP-unit@.service
+
+Update unit
+
+  $ inagoctl --fleet-endpoint=${FLEET_ENDPOINT} update $GROUP --max-growth=0 --min-alive=0 --ready-secs=5
```
xh3b4sd: Why ready secs 5 ? IIRC I recently saw another test where it was 3 or 2.

Nesurion: There is currently no other tests that sets the `ready-secs` flag to something other than the default (30).
Setting it to 5 is arbitrary value less than 30. No particular reason, more a gut feeling.

xh3b4sd: Good I don't trust anybody. https://github.com/giantswarm/inago/commit/45ea56562f031bde35ac18847d8664e16c8afc3f#diff-3c62d059684b69b3b81228eb8f747593R22

Nesurion:
@xh3b4sd RFR

xh3b4sd:
LGTM

PR #248, opened 2016-05-18
Add note about statically compiled kubectl.

apenney:
The kubectl you grab from kubernetes will look in /etc/hosts to resolve
localhost, despite the fact that it lives in /usr/share/baselayout/ on
CoreOS.

For demo purposes a symlink should be suitable.

### DIFF
--- a/docs/k8s.md
+++ b/docs/k8s.md
@@ -46,7 +46,7 @@ And we're done!
 
 ## Testing your Kubernetes cluster
 
-You can check if your cluster is running with [`kubectl`](https://coreos.com/kubernetes/docs/latest/configure-kubectl.html):
+You can check if your cluster is running with [`kubectl`](https://coreos.com/kubernetes/docs/latest/configure-kubectl.html).  If you download the statically compiled version directly from kubernetes you will need to symlink the hosts file into /etc with `sudo ln -s /usr/share/baselayout/hosts /etc/hosts`:
 
 ```
 $ kubectl cluster-info


marians:
Hey @apenney, thanks very much for the contribution!

I can't judge on the content and will leave that to others. But I have a formal comment: Please also set the path `/etc` in backticks, since it's a literal that should stand out from normal text.

Cheers

docs/k8s.md
```diff
@@ -46,7 +46,7 @@ And we're done!
 
 ## Testing your Kubernetes cluster
 
-You can check if your cluster is running with [`kubectl`](https://coreos.com/kubernetes/docs/latest/configure-kubectl.html):
+You can check if your cluster is running with [`kubectl`](https://coreos.com/kubernetes/docs/latest/configure-kubectl.html).  If you download the statically compiled version directly from kubernetes you will need to symlink the hosts file into /etc with `sudo ln -s /usr/share/baselayout/hosts /etc/hosts`:
```
zeisss, line 49: that would mean we overwrite `/etc/hosts`\- I am not sure thats a good idea. It might be more feasible to append it to the end of `/etc/hosts`.

apenney, line 49: At least for me, on coreos (beta), there is no /etc/hosts.

hectorj2f:
There is not much activity in this PR, so I'll close cause it doesn't bring any value atm. Thanks for your contribution.

Issue #250, opened 2016-09-09
Replace file-system with afero


kopiczko:
Hi,

I'm working on #55.
file-system package seems to be feature-incomplete. I miss Stat function for example.
There is a great https://github.com/spf13/afero which is basically file system abstraction.

I'm happy to provide a PR. Please let me know your thoughts.

PR #251, opened 2016-09-09
replace file-system with afero

kopiczko:
Solves #250


teemow:
thx! feels good to remove that code :+1:

lgtm

PR #254, opened 2016-09-10
make go vet happy

kopiczko:
Fix composite literal uses unkeyed fields.

See: https://travis-ci.org/giantswarm/inago/jobs/158933785


hectorj2f:
lgtm

PR #256, opened 2016-09-10
fix extend with content test

kopiczko:


hectorj2f:
This PR has a change which seems to be related to this PR https://github.com/giantswarm/inago/pull/258. Why don't you include it to the same PR ?

kopiczko:
Hmm, it's already included. I wanted to make a few smaller PR's, but I was blocked on #258 so I decided to make a bigger one. I'm fine with closing this in favour of #258.

PR #258, opened 2016-09-17
Pass directory path instead of group name

### LINKED ISSUE #55: support storing groups in parent folders
with a folder structure like:

```
example
├── ser1
│   └── ser1-unit@.service
└── ser2
    └── ser2-unit@.service
```

inago does not support `inagoctl submit example/ser1`. From a usability perspective it would be nice to support parent folder to store groups in categories and not have to `cd` into the parent folder.

kopiczko:
Fixes #55 
It also moves handling #210 to cli package which seems right.

Besides _submit_ this change affects _up_, _update_, and _validate_ commands.

I had to drop FS abstraction at all to use Abs function. It resulted in much less dependencies.


kopiczko:
~~Don't merge yet~~ (fixed). This breaks up command, as submit expects a group's directory path and start expects the group name.

kopiczko:
_up_ command fixed.

zeisss:
is travis misbehaving? :(

JosephSalisbury:
builds should be good now.

kopiczko:
Hmm, I don't see the PR here: https://travis-ci.org/giantswarm/inago/pull_requests
I guess you have PR building enabled since there is this CI check. Something strange is going on.

JosephSalisbury:
Maybe push a new commit to try and kick Travis? I agree, something is funny here

kopiczko:
Stale :)

